PR #1, opened 2015-03-18
Update .travis.yml

Suprcode:


GorzenDev:
could you please explain a bit what Travis actually does.
is it some sort of system integrated in github?
i saw travis give a message that Readme.md could not be build on my last commit in Creature_system

Suprcode:
I've turned off travis builds whilst i try and figure it out on this branch.

Travis CL are basically automated builds. So every commit / change / branch merge, a program will be notified and attempt to compile the program. If it's successful it should (i hope) provide the binary files.

travis-ci.org/

It builds the solution / project on linux, in a program called mono (.net compiler) - and everything you need to tell it to compile needs to go in the .travis.yml file

I'm currently having trouble with it getting external resources - as i expect linux doesn't contain a lot of the Microsoft libraries used.

GorzenDev:
i understand now.
so it actual does create binaries.
thanks for explaining.

PR #2, opened 2015-03-23
Creature system

GorzenDev:
There are some things im unsure if i used the best approach for it.
If you guys can check out the creature expire code and have some feedback on it that would be much appreciated.
Currently a creature's expiretime only decreases when a player is online.(even when unsummoned)
Server.MirObjects.PlayerObject.RefreshCreaturesTimeLeft()


Suprcode:
i think that's correct - a person shouldn't really be punished for only being able to play a short amount of time.

Just like buffs only count down when you're logged in - i'd expect pets to do the same

PR #4, opened 2015-04-06
Added Reincarnation

astrok100:
Hopefully what I coded is okay.
This skill should be 100% working.
Flinching the caster cancels the skill. (currently client tells server)
Skill now levels,
Cast time depending on skill level.
Failure to revive depending on skill level.


Suprcode:
There were some other issues with this spell which needed solving. Such as the "revive in town" message needed to reappear if the user declined reviving.

I'll accept this merge, however some aspects may still need looking at for this spell.

Refer to this thread for details on old bugs
http://www.lomcn.org/forum/showthread.php?88597-Revision-114/page2&highlight=reincarnation

PR #21, opened 2015-04-20
Prevent ProcessPoison NullPointerException

RyanTaplin1705:


RyanTaplin1705:
Personally i'd prefer to find out what is causing the nullpointer and fix that rather than just stopping it from crashing when there is a null pointer. However I can't replicate the problem.

PR #43, opened 2015-07-14
Break on Death added.

UrbanCabbage:
Signed-off-by: Elijah <ash@audis.co.uk>

### DIFF
--- a/Common.cs
+++ b/Common.cs
@@ -2086,6 +2086,7 @@ public class ItemInfo
     public ushort HP, MP;
     public sbyte AttackSpeed, Luck;
     public byte BagWeight, HandWeight, WearWeight;
+    public bool BreakOnDeath;
 
     public bool StartItem;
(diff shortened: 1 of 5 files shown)


Suprcode:
The actual break code looks fine to me, however I was thinking of a better approach to the binds.

If you look in common.cs theres an enum "BindMode"

Ideally this should be updated from byte to short - then the bind modes should be added here. There are also existing binds of "BindOnEquip" and "BindNoSRepair" that should now be moved to this bind enum.

Doing it this way just keeps it tidy with the bind modes all in one place. There's quite a lot to change so just let me know if you don't fancy doing it.

I expect versioning will need to be used to gracefully convert the byte bind to the short bind too.

UrbanCabbage:
I'll take a look, a new area for me to look into. I'll get back to you when I've figured out if its something i can get my head around.

PR #44, opened 2015-07-14
Flags added for NoSpecialRepair, BindOnEquip and BreakOnDeath

UrbanCabbage:
Also added hover effects for all three of them on items.

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -6373,6 +6373,48 @@ namespace Client.MirScenes
 
             #endregion
 
+            #region DONT_SPECIALREPAIR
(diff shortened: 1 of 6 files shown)


Suprcode:
looks good to me.

Only other thing that needs doing is now deleting the 2 old bind variables in ItemInfo that are no longer in use. I'll do that if you like, when you delete them it'll throw a few errors and more version checks will need to be added around them.

UrbanCabbage:
I tried doing this and failed, so yes, if you wanna delete and fix it, i'll probably look over code when it's done so i can see how you did it.

Just did a new pull request to fix a couple things i'd left over.

PR #45, opened 2015-07-14
* When the player appears on the map, to modify the monster does not …

msh0411:
When the player appears on the map, to modify the monster does not attack the problem directly.


Suprcode:
well spotted. Didn't cross my mind that it won't check again for a minute.

thanks.

PR #48, opened 2015-07-17
Monster

Stupple:
all rest monster added top to 402 

402 this bunny 2 witch was 2 monster in one lib


Client/MirObjects/MonsterObject.cs
```diff
@@ -2108,6 +2779,18 @@ public override void DrawEffects()
                             break;
                     }
                     break;
+                    //stupple may need to take out ???
```
Suprcode: why do these need to be taken out?

Server/MirEnvir/Envir.cs
```diff
@@ -19,7 +19,7 @@ public class Envir
         public static object AccountLock = new object();
         public static object LoadLock = new object();
 
-        public const int Version = 53;
+        public const int Version = 55;
```
Suprcode: why was the version increases for this request? I don't see anything that needs it

Suprcode:
This request mostly looks good. Chaosdeath has provided names for every monster - but once i've accept this I will go through an update them.

The "bunny2" that was added at 402 might as well go in an empty slot (like at 397). I can change that after though in my set.

If you can just answer the 2 issues - and also update the version number back to 53 if you can.

Client/MirObjects/MonsterObject.cs
Stupple: this put ring round tho monster it USA mir did. i cant remember if the right image if i cord this right i change it few time

Server/MirEnvir/Envir.cs
Stupple: no idea this wast me i notice it on my log as but i didn't sent this only thing was set to 55 was hum up

Stupple:
here was monster i taked out my need to replace it at end and put new frame set number. no idea why he was add to end after it was all rdy coded

Common.cs
```diff
@@ -359,13 +359,169 @@ public enum Monster : ushort
...
+    Sanke15 = 258,
+    Sanke16 = 259,
+    Sanke17 = 260,
+
+    DeathCrawler = 261,
+    BurningZombie = 262,
+    MubZommbie = 263,
+    FrozenZombie = 264,
+    UndeadWolf = 265,  // mountain pass same mobs
+    Demonwolf = 266,   // mountain pass same mobs
```
Suprcode: what did you mean by this comment? "// mountain pass same mobs"

Stupple: nothing it just note for when i added Ai and so on

PR #50, opened 2015-07-20
Healingfield

Stupple:


Stupple:
also note this was done before you change the cool down timers so it may need be added some where els

Suprcode:
This isn't how healing field works officially.

It should be a spell object - which you cast on the floor than becomes like a temporary "safe zone" that you can walk in and out of and so can others to get a heal.

But all i really wanted to see was if you did anything different with the drawing. Can't see why mine looked so bad.

Stupple:
o.O didnt no that but i add it as flaming field but with healing and biase heal with area of ffs  it good tho heal like mass healing. also put ta at rish be in close range and make them more off assist in lvling

PR #78, opened 2015-08-26
Auto patcher

ElijahLOMCN:


Suprcode:
change namespace to launcher from autopatcher

PR #111, opened 2015-10-03
Gem Independence

Duane01:
Gem stat independence via config. If false gemming success is based on total gem count. Otherwise it is based on the relevant stat.

Add "GemSystem.ini" to config files.

Content:

[Config]
GemStatIndependent=True


Server/MirObjects/PlayerObject.cs
```diff
@@ -10842,16 +10951,101 @@ private bool ValidGemForItem(UserItem Gem, byte itemtype)
...
+            else if ((gem.Info.MP + gem.MP) > 0)
+                return StatType.MP;
+
+            else if ((gem.Info.HealthRecovery + gem.HealthRecovery) > 0)
+                return StatType.HP_Regen;
+
+            // These may be incomplete. Item definitions may be missing?
+
+            else if ((gem.Info.HPrate) > 0)
+                return StatType.HP_Precent;
```
Suprcode: HP_Precent should be HP_Percent

Server/MirObjects/PlayerObject.cs
```diff
@@ -10842,16 +10951,101 @@ private bool ValidGemForItem(UserItem Gem, byte itemtype)
...
+
+            else if ((gem.Info.MPrate) > 0)
+                return StatType.MP_Precent;
```
Suprcode: MP_Precent should be MP_Percent

PR #189, opened 2016-10-17
Allow pet poison to be recognised

RyanTaplin1705:
1. Check monster is a pet.
2. Check the pet master is current attacker
3. Assign currentAttacker to pet master (player)
4. Send message to player with variation in text based on it being pet or player poison.


RyanTaplin1705:
I didn't test this however I assume it's all ok. Looks like I caught all the player objects necessary.

PR #191, opened 2016-10-17
Fixed #190 GuildStore grey box issue.

### LINKED ISSUE #190: GuildStorage causing greybox
My assumption is that it is locking the blocks when initiating the transfer however it doesn't unlock the boxes when the transfer fails (e.g. due to being outside safezone).

I will investigate and look for the solution in due time if none is provided.

http://www.lomcn.org/forum/showthread.php?95007-Guild-Storage

RyanTaplin1705:


Suprcode:
type 2 is a request list - however you've changed it on the move item from guild storage to guild storage.

This should actually be type 3, and a new server code should be written to allow items to be moved from one cell to another.

RyanTaplin1705:
Ah yeah I see, I missed requestlist.

Surely for ease it would be better to move move request list further down in the switch statement and move everything else up one server side. That requires 2 little numbers to be changed.

RyanTaplin1705:
@Suprcode before I made this change;

Type=3 to Type=2 (This following piece of code wasn't being used anywhere).
[code]
MyGuild.SendServerPacket(new S.GuildStorageItemChange() { Type = 2, User = Info.Index, Item = MyGuild.StoredItems[to], To = to, From = from });
[/code]

So in effect this code is never being called?
[code]
        private void GuildStorageItemChange(S.GuildStorageItemChange p)
        {
            MirItemCell fromCell = null;
            MirItemCell toCell = null;
            switch (p.Type)
            {
case 2:
                    toCell = GuildDialog.StorageGrid[p.To];
                    fromCell = GuildDialog.StorageGrid[p.From];

                    if (toCell == null || fromCell == null) return;

                    toCell.Locked = false;
                    fromCell.Locked = false;
                    fromCell.Item = toCell.Item;
                    toCell.Item = p.Item.Item;
                    
                    Bind(toCell.Item);
                    if (fromCell.Item != null) 
                        Bind(fromCell.Item);

                    break;
[/code]

Suprcode:
but my point is theres no server side code to move the items from cell to cell - if you look at the 3 types available, one is remove from guild, one is add to guild, the other is refresh list.

There is no case statement to move from cell to cell in the server - only clientside.

So for this to work you would need to add the server code to go along side the existing client cell to cell move code which you pasted above.

RyanTaplin1705:
I'm sure this is already being done?! In Server.MirObjects.PlayerObject

```

                case 2: // Move Item
                    GuildStorageItem q = null;
                    if (!MyGuildRank.Options.HasFlag(RankOptions.CanStoreItem))
                    {
                        Enqueue(p);
                        ReceiveChat("You do not have permission to move items in guild storage.", ChatType.System);
                        return;
                    }
                    if (from < 0 || from >= MyGuild.StoredItems.Length)
                    {
                        Enqueue(p);
                        return;
                    }
                    if (to < 0 || to >= MyGuild.StoredItems.Length)
                    {
                        Enqueue(p);
                        return;
                    }
                    if (MyGuild.StoredItems[from] == null)
                    {
                        Enqueue(p);
                        return;
                    }
                    if (MyGuild.StoredItems[from].Item.Info.Bind.HasFlag(BindMode.DontStore))
                    {
                        Enqueue(p);
                        return;
                    }
                    if (MyGuild.StoredItems[to] != null)
                    {
                        q = MyGuild.StoredItems[to];
                    }
                    MyGuild.StoredItems[to] = MyGuild.StoredItems[from];
                    if (q != null) MyGuild.StoredItems[from] = q;
                    else MyGuild.StoredItems[from] = null;

                    MyGuild.SendItemInfo(MyGuild.StoredItems[to].Item);

                    if (MyGuild.StoredItems[from] != null) MyGuild.SendItemInfo(MyGuild.StoredItems[from].Item);

                    MyGuild.SendServerPacket(new S.GuildStorageItemChange() { Type = 2, User = Info.Index, Item = MyGuild.StoredItems[to], To = to, From = from });
                    MyGuild.NeedSave = true;
                    break;

```

Suprcode:
ah yes, i guess i was half awake last night and didn't scroll all the way down.

with that said, is this commit suitable to be accepted?

RyanTaplin1705:
I think so? I didn't see anything wrong other than the above **case2** now **case3** code not being used anywhere (to my knowledge) which will need to be removed; could also refactor the algorithm type+3 to type+2 and shift all case statements down by 1 otherwise you have a gap case0;case1;case3;case4;case5.

Not sure it's a big issue though.

I assume some of the stuff could be done nicer?! I've not really looked into that too much yet; only looked at this issue at hand.

PR #193, opened 2016-10-18
Stop ridiculously long IC names

RyanTaplin1705:
Fix for this IntelligentCreature injection; I was going to add a regex to this (it doesn't / shouldn't harm the server) when injecting ASCII characters. If someone has injected ASCII characters then you can identify something fishy going on.

http://www.lomcn.org/forum/showthread.php?94317-UpdateIntelligentCreature

Ideally need to create a server side save for createRenameEnabled and enable it in the following piece of code however i'm currently on the MySQL source so it will require an additional column to be added and populated (same for flat file).

`                            case 20://Mirror
                                Enqueue(new S.IntelligentCreatureEnableRename());
                                break;
`

### DIFF
--- a/Server/MirNetwork/MirConnection.cs
+++ b/Server/MirNetwork/MirConnection.cs
@@ -1569,7 +1569,8 @@ namespace Server.MirNetwork
                 {
                     if (Player.Info.IntelligentCreatures[i].PetType == petUpdate.PetType)
                     {
-                        Player.Info.IntelligentCreatures[i].CustomName = petUpdate.CustomName;
+			String newName = petUpdate.CustomName;
+                        Player.Info.IntelligentCreatures[i].CustomName = newName.Length <= 12 ? newName : Player.Info.IntelligentCreatures[i].CustomName;
                         Player.Info.IntelligentCreatures[i].SlotIndex = petUpdate.SlotIndex;
                         Player.Info.IntelligentCreatures[i].Filter = petUpdate.Filter;
                         Player.Info.IntelligentCreatures[i].petMode = petUpdate.petMode;
--- a/Server/MirNetwork/MirConnection.cs
+++ b/Server/MirNetwork/MirConnection.cs
@@ -1569,8 +1569,8 @@ namespace Server.MirNetwork
                 {
                     if (Player.Info.IntelligentCreatures[i].PetType == petUpdate.PetType)
                     {
-			String newName = petUpdate.CustomName;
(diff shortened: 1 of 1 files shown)


Suprcode:
your change simply replaces the name with the current name if its too long - is there already suitable client checks to stop this happening in the first place?

Also theres no real need for the temporary variable and resaving over the name. i would just do

if(petUpdate.CustomName.Length < 13)
Player.Info.IntelligentCreatures[i].CustomName =  petUpdate.CustomName;

RyanTaplin1705:
Yes there is suitable client side checks. I think the same thing which checks your textboxInput is < 20 however players can modify the client and inject insanely long / stupid data as shown on LOMCN thread.

I did have a SS of this; it was taking up like 50% of the screen with spammed ASCII characters at one point went I went overboard.

Yes that's my bad, I originally added the Regex so it made more sense t make a string and assign the temp name to it instead of constantly using petUpdate.CustomName however I decided to take it out as it require importing the Regex dependencies which I didn't know was worth doing.

`//CreatureL.CustomName = "❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n❚█══█❚♥♥♥♥♥♥♥♥♥♥♥♥♥(っ◕‿◕)っ♥♥♥♥♥♥♥♥♥♥♥♥♥❚█══█❚\n";`

PR #207, opened 2016-10-19
Typo 2 in IC

RyanTaplin1705:
testing pull req from GUI


RyanTaplin1705:
wah didn't want that to happen

PR #217, opened 2016-11-10
Patches

RyanTaplin1705:
Will this ever be implemented? Maybe move it to suggestion section or something.
http://www.lomcn.org/forum/showthread.php?94834-Bug-RefreshEquipmentStats

May need closing?
http://www.lomcn.org/forum/showthread.php?94842-Expiry-items

Needs Closing:
http://www.lomcn.org/forum/showthread.php?94635-Taoist-(Mass)Hiding-breaks-on-Physical-Attack
http://www.lomcn.org/forum/showthread.php?94317-UpdateIntelligentCreature
http://www.lomcn.org/forum/showthread.php?94493-npc-quot-Removeskill-quot-or-quot-Deleteskill-quot
http://www.lomcn.org/forum/showthread.php?94321-Flashdash-and-MagicShield

Couldn't replicate [DeaD SeRiOuS](http://www.lomcn.org/forum/member.php?27181-DeaD-SeRiOuS) error in regards to FD and MS. However I added MagicShield effect as posted by petesnoobeh

### DIFF
--- a/Client/MirControls/MirItemCell.cs
+++ b/Client/MirControls/MirItemCell.cs
@@ -1113,17 +1113,13 @@ namespace Client.MirControls
                         {
                             case MirGridType.GuildStorage: //From Guild Storage
                                 if (GameScene.SelectedCell.GridType == MirGridType.GuildStorage)
-                                {
-                                    //if (Item != null)
-                                    //{
-                                        //GameScene.Scene.ChatDialog.ReceiveChat("You cannot swap items.", ChatType.System);
-                                        //return;
-                                    //}
+                                {                                    
                                     if (!GuildDialog.MyOptions.HasFlag(RankOptions.CanStoreItem))
                                     {
                                         GameScene.Scene.ChatDialog.ReceiveChat("Insufficient rights to store items.", ChatType.System);
                                         return;
                                     }
+                                       
                                     //if (ItemArray[ItemSlot] == null)
                                     //{
                                         Network.Enqueue(new C.GuildStorageItemChange { Type = 2, From = GameScene.SelectedCell.ItemSlot, To = ItemSlot });
--- a/Client/MirObjects/PlayerObject.cs
+++ b/Client/MirObjects/PlayerObject.cs
@@ -650,6 +650,18 @@ namespace Client.MirObjects
 
             if (RidingMount) return;
 
+            if (CurrentEffect == SpellEffect.MagicShieldUp)
+            {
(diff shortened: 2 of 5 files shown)


RyanTaplin1705:
@Suprcode faralaralar you need to be on call when I make merge requests 😝 🤓

Suprcode:
I've already made a crafting system which works well, but i added it on another source i work on. Uses recipes generated from text files and alters the buy function to require all recipe items instead of needing gold, works very smoothly and is a non complicated system unlike the official one. Will probably add it in the main source eventually. If you want to add a crafting system then please do as it's unlikely i'll think about merging it until after xmas, obviously i'll not merge it unless its a complete working system.

RyanTaplin1705:
Damn I thought you was going on holiday or something like permanently? 😛 

Mine will probably never be as good; i'm surprised you did it through text files though.

I was thinking of having like ... (here is a mock) although I don't know if its worth me stepping on your toes if it's going to be overwritten. Is there something in particular from the [TODO](http://wiki.mironline.co.uk/Todo) that you haven't done? Most likely going to tell me monsters? 😆 

untitled-2
untitled-5

RyanTaplin1705:
@Suprcode hmm how about if I tried to implement a element system? Would you incorporate it into the source?

Suprcode:
I added it through text files as they're a perfect fit for recipes - the file name has to be unique (as to item names) and the recipes require a single line and quantity.

The exact same reason drop file are done as text files and not database stored.

I think anything more is unnecessarily complex for what a crafting system needs to be, as you can achieve exactly what you've got in your screenshots inside a basic textfile. 

Requirement, level, cost, quantity are already used with the ItemInfo and the Item DB - theres no need to duplicate it. So the only thing left is the recipe, which can be easily stored in a text file.

Suprcode:
I doubt i would merge a custom feature like an element system in to the main source, however don't let that stop you from doing it.

RyanTaplin1705:
Interesting approach. It makes a lot of sense. So in this case your item cost will be your crafting cost ? (if you've chosen to have a crafting fee)

I feel that if I do something for myself it will end up getting lost and I'll end up deleting it or something along those lines. Well we'll see if I pick it up; I think having a element system for spells and an element defense system would be a good way for providing a diverse skill usage in game. Instead of using primarily flame disruptor the spell will match the cave that the user will hunt in (e.g. fire for frosttiger, ice for minotaurking etc).

Issue #221, opened 2016-11-18
Damage display


RyanTaplin1705:
If you miss and crit it comes up with both. Should just bypass the crit message if it misses.

PR #222, opened 2016-11-19
Can't be a crit if damage is <= 0

RyanTaplin1705:
Fix issue #221


Suprcode:
why are you sending the critical packet at all if you're not going to display it? It would be more logical to not send the packet and correct the issue at server side.

RyanTaplin1705:
Sure my error. I'vemMoved the crit packet/check to below the armour check; 
It will never reach crit packet broadcast if it sends a miss packet due to returning out of method with 0.

PR #223, opened 2016-11-19
Prevent users from buying more than they can retreive.

RyanTaplin1705:


Suprcode:
i dont understand these changes. why are you limiting purchase quantity to 99, and what is this line trying to do?

if (((decimal)(Quantity * Product.Count) / Product.Info.StackSize) > 5) return;

why 5?

RyanTaplin1705:
Following same format as front end [code](https://github.com/Suprcode/mir2/blob/613cd0d50bbd18050e7d3562a86dd1d871510179/Client/MirControls/MirGameShopCell.cs#L156)

```
                if (((decimal)(Quantity * Item.Count) / Item.Info.StackSize) > 5) Quantity = ((5 * Item.Info.StackSize) / Item.Count) > 99 ? Quantity = 99 : Quantity = (byte)((5 * Item.Info.StackSize) / Item.Count);
                if (Quantity >= 99) Quantity = 99;
                if (Item.Stock != 0 && Quantity > Item.Stock) Quantity = (byte)Item.Stock;
```
Also the stack size has to be **<= 5** because the mail box can only hold stacks of 5 I assume.

RyanTaplin1705:
Decided what to do with this yet? I think this needs to be done server side however maybe remove the 99 cap? It will then just be a 255 cap.

Suprcode:
99 cap is because ui only supports up to 99 so is ok. will accept when im
next at a pc

RyanTaplin1705:
Ok cool, by the way I think the UI fits 3 digit numbers fine. 

untitled

ElijahLOMCN:
Don't be fucking with my game shop you cunts :)

RyanTaplin1705:
Do a better job next time then @ElijahLOMCN 😂 joking.

ElijahLOMCN:
Game shop is the most comprehensive feature in the game I often suggested we call it Gameshop and have one of its features as 'Play Mir while you shop'

lodos3:
I designed it !!

lodos3:
How about one of you get text fields to not render above the UI? That'd be
an idea

RyanTaplin1705:
@lodos3 Not really looked into it. Jamie done it in his marble files and said it was a fairly difficult task.

lodos3:
Wouldn't it be logical to just disable the code for text field unless the
area is selected, so that if anywhere BUT that form area is selected on the
Dialog or any other dialog, it would remove the text field. I'm certain
that all the designs have a Graphical Placement for each text field, so
there is really no need for the text form to be on all the time.

Suprcode:
easier said than done. it's a rewrite of the entire text control to stop it drawing above everything.

PR #232, opened 2017-04-30
Server side interface updates

Combie81:
Name change
Version date update
Launcher target website changed from non working one to LOMCN
Server config and text changes
Add new delay label to identify minutes
Increase monster stat box length and also add validation to the
fields(0-65535)
Rearrange box layout
Added lottery ticket handling


Suprcode:
thanks for updates - i've changed the broken launcher link to the updated one (instead of lomcn) - it's a little outdated but points people to the relevant sections on the wiki and lomcn

https://launcher.mironline.co.uk/web/

PR #233, opened 2017-05-01
TrustMerchant

daneo1989:
New TrustMerchant updated.


daneo1989:
image 1
image 2

New Title.lib
https://www.dropbox.com/s/69lz7s0czrwyyfq/Title.Lib?dl=0

daneo1989:
Updated, Basically not able to have multiple Network.Enqueue(new C.MarketSearch in one case, 

Reverted back to have 1 search per case, due to this bug (You couldn't select IE Drapery Items category and view Armour Helmet Belts Boots and Stones)

I have reverted it back so I can spend more time looking into having multiple Type = ItemType. in the same case search. 
image 4

PR #245, opened 2017-06-03
Conquest Wall Respawn fix

daneo1989:
Conquest Wall Respawn fix provided by Cookie on LOMCN


Suprcode:
there is no fix provided in this pull request.

PR #247, opened 2017-06-15
Fixed errors

brandc87:
Fix for assembly version and Library Editor.


Suprcode:
added the missing files to the project so will cancel this request

PR #251, opened 2017-07-10
Archer base attack stat change

meacher0:
Seemingly allows archer to hit stuff by swapping MC stat for DC.


daneo1989:
Archers was changed to MC from DC

CookieLOM:
Cant see why it needs changing back myself leave it and do stuff worth changing :+1:

meacher0:
my bad

PR #265, opened 2017-12-19
RequiredType update,

daneo1989:
RequiredType update, Can now choose between Min & Max AC,MAC,DC,SC,MC

### DIFF
--- a/Common.cs
+++ b/Common.cs
@@ -2097,7 +2097,7 @@ public static class Globals
         MaxCharacterNameLength = 15,
         MaxCharacterCount = 4,
 
-        MaxChatLength = 115,
+        MaxChatLength = 80,
 
         MaxGroup = 15,
         
@@ -2117,8 +2117,8 @@ public static class Globals
 
     public const uint SearchDelay = 500,
                       ConsignmentLength = 7,
-                      ConsignmentCost = 1000,
-                      MinConsignment = 1000,
+                      ConsignmentCost = 5000,
+                      MinConsignment = 5000,
                       MaxConsignment = 50000000;
 
 }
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -56,3 +56,9 @@ Client/Client.csproj
 /AutoPatcherAdmin/obj
 
 /Build
+.vs/Legend of Mir/v15/Server/sqlite3/db.lock
+.vs/Legend of Mir/v15/Server/sqlite3/storage.ide
+.vs/Legend of Mir/v15/Server/sqlite3/storage.ide-shm
+.vs/Legend of Mir/v15/Server/sqlite3/storage.ide-wal
```
Suprcode: these sql references and files need removing from the source

Server/Server.csproj
```diff
@@ -59,6 +59,7 @@
     <WarningLevel>4</WarningLevel>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <Prefer32Bit>false</Prefer32Bit>
+    <LangVersion>7.1</LangVersion>
```
Suprcode: does the language need to be 7.1? Unless there is specific syntax which requires 7.1 it should use the lowest possible. Otherwise you're forcing every coder to unnecessarily update

daneo1989: I basically changed to this version so it would compile think it had problems compiling on default for some reason

Common.cs
```diff
@@ -967,11 +967,16 @@ public enum RequiredGender : byte
...
+    MaxAC = 2,
+    MinMAC = 3,
+    MaxMAC = 4,
+    MinDC = 5,
+    MaxDC = 6,
+    MinMC = 7,
+    MaxMC = 8,
+    MinSC = 9,
+    MaxSC = 10,
 }
```
Suprcode: This will break everyones DB - old SC value of 5 is now MinDC - so all types on items will be wrong.

You need to put any new values at the end, enums which are numbers shouldn't be reordered

PR #268, opened 2018-02-09
Server Performance Change, bug fix and a value change

Pete107:
ItemInfo will no longer be sent when a Player logs in-game.
Instead the ItemInfos will load from a local Binary File which you can export using the ItemInfo Form on the Server, more info provided on LOMCN.
SMain Form opening bug fix.
Check Level now uses ushort.

### DIFF
--- a/Client/MirControls/MirScene.cs
+++ b/Client/MirControls/MirScene.cs
@@ -198,7 +198,28 @@ namespace Client.MirControls
 
         private void NewItemInfo(S.NewItemInfo info)
         {
-            GameScene.ItemInfoList.Add(info.Info);
+            #region Old ItemInfo loading (1 by 1) Pete107 Edens Elite
+            //GameScene.ItemInfoList.Add(info.Info);
+            #endregion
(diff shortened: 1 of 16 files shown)


Pete107:
Should be the last change required, the .dat file is automatically created and updated just as the normal DB would.
Removed the export button, removed the export code in ItemInfo form.
The Walk logic is in there so don't forget if you plan on keeping it this awful (my opinion) I just prefer the way it was when I enjoyed playing.

CookieLOM:
This may be a dumb question but how is the walk logic in here better than standard in a simple terms ha and could you point out the area which these changes are in as although the rest of the code looks great I am happy with the original setup as I've highly changed it anyway

Pete107:
If you played euro or any server before Crystal files you'll notice when you run head on into a tree it'll 'redirect' you to the next direction that is available to move to (only 3 directions in total, the direction you face, the previous and the next direction.

https://github.com/Suprcode/mir2/pull/268/files#diff-3d78b08a721ba384536a10b85b899f1fR9365

PR #273, opened 2018-03-02
Public Events

H4nySam:
Server
Database Impact/Changes
- Upped CustomVersion to 1. (Any of the below changes gets read if custom version >= 1 , if your using your own customversion it may conflict)
- New Field MapEventIndex (Gets Saved with Main DB File)
- New Field to MapInfo List<PublicEventInfo>
- 2 New fields to CharacterInfo (DailyEventsCompleted,WeeklyEventsCompleted)
- New DB File (Server.MirRTVDB) used to store and read runtime values such as (CurrentWeekOfTheYear,CurrentDayOfTheYear)

Logic Impact/Changes
- Map Now holds List<Events> and events gets processed within map process (Responsible of activating events)
- New Logic (MapEvent) - Events logic , deactivation,personal loot etc
- RespawnInfo - Added a new field IsEventObjective only used in runtime doesn't get saved/loaded, it gets set from Events logic when activating/spawning monsters
- MonsterObject Monsters in general will act different if its respawn is EventObjective (Events logic retreat,register contributors to event etc)
- PlayerObject Walk,Run should detect if you moved into an event and it will send you a packet to display the event dialog, Server Packets GetMapInformation & MapChanged now holds List<MapEventClientSide> to display events in the client
- Envir On server boot and every one hour server runs checks to clear players daily & weekly completed events (You can add currencies there aswell if your files has extra currencies like paragon mir and you want to implement some sort of a cap)

Client
- New field to MapControl (List<MapEventClientSide>) it gets cleared and refilled when MapChanged/MapInformation packets are receive.
- New Dialog Event Dialog Appears when you enter an event (Shows progress of the current event)
- BigMapDialog Displays List<MirControlImage> within on-draw event, it represents current map events, it uses Items.Lib, each event type choses a different image you can customize this based on your needs.

### DIFF
--- a/Client/Client.csproj
+++ b/Client/Client.csproj
@@ -189,6 +189,7 @@
     <Compile Include="MirScenes\Dialogs\BuffDialog.cs" />
     <Compile Include="MirScenes\Dialogs\ChatNoticeDialog.cs" />
     <Compile Include="MirScenes\Dialogs\ChatOptionDialog.cs" />
+    <Compile Include="MirScenes\Dialogs\EventDialog.cs" />
     <Compile Include="MirScenes\Dialogs\FishingDialog.cs" />
     <Compile Include="MirScenes\Dialogs\FriendDialog.cs" />
     <Compile Include="MirScenes\Dialogs\GameshopDialog.cs" />
--- a/Client/MirGraphics/MLibrary.cs
+++ b/Client/MirGraphics/MLibrary.cs
@@ -46,7 +46,9 @@ namespace Client.MirGraphics
         public static readonly MLibrary
             Items = new MLibrary(Settings.DataPath + "Items"),
             StateItems = new MLibrary(Settings.DataPath + "StateItem"),
-            FloorItems = new MLibrary(Settings.DataPath + "DNItems");
+            FloorItems = new MLibrary(Settings.DataPath + "DNItems"),
+            StateEffect = new MLibrary(Settings.DataPath + "StateEffect");
+
 
         //Deco
         public static readonly MLibrary
--- a/Client/MirObjects/ItemObject.cs
+++ b/Client/MirObjects/ItemObject.cs
@@ -7,6 +7,7 @@ using Client.MirGraphics;
 using Client.MirScenes;
 using S = ServerPackets;
 using System.Text.RegularExpressions;
+using Client.MirControls;
 
 namespace Client.MirObjects
 {
@@ -23,9 +24,21 @@ namespace Client.MirObjects
 
         public Size Size;
 
+        public MirAnimatedControl _effect;
+        public ItemGrade Grade;
 
         public ItemObject(uint objectID) : base(objectID)
         {
+            _effect = new MirAnimatedControl()
+            {
+                Animated = true,
+                AnimationCount = 25,
+                AnimationDelay = 100,
+                Index = 680,
+                Library = Libraries.StateEffect,
+                Loop = true,
+                Blending = true,
+            };
         }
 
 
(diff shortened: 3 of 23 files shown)


daneo1989:
Thank you for the release Hany, couple questions, bugs, suggestions

Bug: Event spawned Monsters, can venture and be killed outside Event area,
Suggestion: Monsters cannot leave Event area, cannot be attacked or killed while stood outside the EA.
Any persons who are not in EA cannot gain EXP, Gold rewards from killed monsters.

Question: Monster Spawns, If you put Spread 10 on X,Y the monsters will just spawn at 10:10, shouldn't this spread across all 0-10 X,Y rather than just one X,Y?
A possible exploit could be that people will nowhere the Boss would spawn which could give guilds the possibility to own all events.

Suggestion: The Objective list, wouldn't this suffice with being on an Idx list, rather than open like it is, 
Maybe possibly having a new field of Objective Message, (Basically information on what to do)

H4nySam:
You're most welcome daneo,

> Bug: Event spawned Monsters, can venture and be killed outside Event area,

I felt i should leave it like that, since i have forced the personal loot bit to only be given to players within the actual zone, + with retreat feature implemented at 10 steps i thought it would be too extreme tbh'


> Suggestion: Monsters cannot leave Event area, cannot be attacked or killed while stood outside the EA.
> Any persons who are not in EA cannot gain EXP, Gold rewards from killed monsters.

I like it , maybe even make the retreat functionality in the monster to be triggered once monster is outside of the EA.

> Question: Monster Spawns, If you put Spread 10 on X,Y the monsters will just spawn at 10:10, shouldn't this spread across all 0-10 X,Y rather than just one X,Y?
> A possible exploit could be that people will nowhere the Boss would spawn which could give guilds the possibility to own all events.

I understand where you're coming from, and actually i've been working on a patrol feature on the spawn, so this spawn will just patrol the event area, going from center to top and to bottom, left and right aswell

> Suggestion: The Objective list, wouldn't this suffice with being on an Idx list, rather than open like it is,
> Maybe possibly having a new field of Objective Message, (Basically information on what to do)
Yeah everything is possible, i really haven't spent much time on this bit but it could do with alot more improvements, I think i followed the worst approach within this bit where i send a comma separated monster names as an objective.

As for objective message it can be either hard coded like how it is atm (Switch on event type and write different text) , or introduce it as a field on event info.


> Once you enter the Event area, Quest type info pops up under the heading EventType,

Do you mean replace the event dialog progress bit with Quest progress style? 
eg
Event Name in green bold text
MonsterName 0/1

Anywho thanks for giving it a try, and thanks for the suggestions they're all noted, i will commit changes soon.

H4nySam:
I get what you mean, but on wow you couldn't know the progress of the event from anywhere you had to move and get within the event to know the progress.

It is possible though will just have introduce new fields to the MapEventClientSide class and modify the 2 bits where MapChanged and MapInfo packets gets sent and populate new fields.

This means all players on the map now will hold a non accurate progress of all events within the map, Progress will only be updated if you teleport/revisit the map etc.

So you will have a problem you will need to update the progress for events every now and then for all players on the map.

Personally i think it's not worth it, you shouldn't know a progress of an event without having to visit it, too much hassle :P 

But i do agree with the quest layout, maybe this dialog is taking too much space..

daneo1989:
might be worth giving the option to have both :)

Pete107:
Within the Attacked method(s) I advise some extra null checks (I think it was Respawn that threw null object error)
https://github.com/Suprcode/mir2/pull/273/files#diff-b9fbad39a736d6bd3170692e1bee855fR2125

https://github.com/Suprcode/mir2/pull/273/files#diff-b9fbad39a736d6bd3170692e1bee855fR2220

H4nySam:
Thanks Pete, i didn't know a respawn would be null if monster spawn was forced.

Commited a fix

PR #282, opened 2018-05-30
 Draw Item Grade Effect

daneo1989:
Huge thanks to Hany for the release of his code,

Draw Item Grade effect to work with Sanjian DropEffects.Lib release on LOMCN
Drop View will disable the effect

### DIFF
--- a/Client/MirObjects/ItemObject.cs
+++ b/Client/MirObjects/ItemObject.cs
@@ -7,6 +7,7 @@ using Client.MirGraphics;
 using Client.MirScenes;
 using S = ServerPackets;
 using System.Text.RegularExpressions;
+using Client.MirControls;
 
 namespace Client.MirObjects
 {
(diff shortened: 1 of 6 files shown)


Client/MirObjects/ItemObject.cs
```diff
@@ -105,6 +119,29 @@ public override void DrawName()
...
+                    case ItemGrade.Legendary:
+                        _effect.BackColour = Color.DarkOrange;
+                        _effect.ForeColour = Color.DarkOrange;
+                        break;
+                    case ItemGrade.Mythical:
+                        _effect.BackColour = Color.Purple;
+                        _effect.ForeColour = Color.Purple;
+                        break;
+                }
+                Libraries.DropEffects.DrawBlend(0 + _effect.Index, new Point(DrawLocation.X, DrawLocation.Y), _effect.BackColour, true, 10F);
```
Suprcode: I'm not sure why you're drawing it manually here? Since you've already created an animated object capable of doing everything, you should be able to just call

_effect .Draw();

Client/MirObjects/ItemObject.cs
```diff
@@ -105,6 +119,29 @@ public override void DrawName()
             NameLabel.Draw();
         }
 
+        public void DrawGradeEff()
+        {
+            if (Grade > ItemGrade.Common)
```
Suprcode: The grade doesn't change within the life of the ItemObject. This would be better being set once inside the Load(S.ObjectItem info) method - instead of being set to the same value every time it needs to update a frame.

Client/MirObjects/ItemObject.cs
```diff
@@ -105,6 +119,29 @@ public override void DrawName()
             NameLabel.Draw();
         }
 
+        public void DrawGradeEff()
```
Suprcode: Can this be renamed to "DrawGradeEffect". No need to shorten method names.

PR #286, opened 2018-10-12
Sockets

RyanTaplin1705:
Some tweaking still needed for GameScene.cs for loading images in. Images also need to be created / added. Recommended size for socket in item label is 15x15.

Sockets can be combined the same as Gems (CTRL + click into item with sockets)

Rune type is available in item database. User can enter stats into the item database and stats will be added to character accordingly. Made the decision to not display rune stats on item label due to it not being feasible to display all stats (if multiple are added) to a rune.

image

~Need to add socket logic into RANDOMSTATS and then added to item drop logic accordingly.~
Sockets added to RIS, now will add sockets on @MAKE and through drops (needs to be configured).

image

### DIFF
--- a/Client/MirObjects/UserObject.cs
+++ b/Client/MirObjects/UserObject.cs
@@ -318,6 +318,34 @@ namespace Client.MirObjects
                 ASpeed = (sbyte)Math.Max(sbyte.MinValue, (Math.Min(sbyte.MaxValue, ASpeed + temp.AttackSpeed + RealItem.AttackSpeed)));
                 Luck = (sbyte)Math.Max(sbyte.MinValue, (Math.Min(sbyte.MaxValue, Luck + temp.Luck + RealItem.Luck)));
 
+                for (int si = 0; si < temp.socketCount; si++)
+                {
+                    ItemInfo sockInfo = temp.sockets[si];
+
+                    if (sockInfo == null) continue;
+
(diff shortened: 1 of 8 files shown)


daneo1989:
Does the Item stats change once a rune is added?

RyanTaplin1705:
@daneo1989 not the item stats itself but the player stats update correctly.

image

Could make it so that it is displayed like normal item adds (+x dc) either coupled with the item/gem added stats or separate so you have (+x dc)(+x dc)

daneo1989:
me personally would need to have the stats displayed on the item, its a good feature, might be a good idea for future to add RuneWords where if you get the RW it gives bonus stats, similar to Sets

RyanTaplin1705:
Yeah I have had that comment from some other people, it's a bit ambiguous when someone has a rune named like I have (i.e testrune1, testrune2) on there item. I was hoping that people would name there runes a bit more distinctively named so that players would easily correlate a rune and its stats to the name of the rune (i.e DCRune and MCRune).. 

I dont' know what you mean by 'RuneWords' but I still have the issue that I dont want to overcrowd the item label, maybe it would be an idea to include bind key (i.e alt or ctrl) that displays extra/hidden information when held and hovered over an item.

daneo1989:
crescentmoon

RyanTaplin1705:
I see, to be honest I didn't like how it ended up. There is a bunch of stuff I would have change but this was just me experiment and digging around in the code a little for a friend.

PR #292, opened 2018-11-13
RespawnInfo Import Fix

RyanTaplin1705:
Auto import is not returning true MonObject Index but instead the index of the MonObject in the MonsterInfoList.

See attached the incorrect debug output. Now fixed with proposed change 👍 
untitled

### DIFF
--- a/Server/MirForms/Database/MapInfoForm.cs
+++ b/Server/MirForms/Database/MapInfoForm.cs
@@ -1582,7 +1582,7 @@ namespace Server
 
                     RespawnInfo respawnInfo = new RespawnInfo
                     {
-                        MonsterIndex = monsterIndex + 1,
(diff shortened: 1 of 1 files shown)


RyanTaplin1705:
I assume this happens when players delete elements in their list. Same cause may effect map and changes will need to happen for the line following this? 

```
 int index = Envir.MapInfoList.FindIndex(a => a.FileName == MirForms.ConvertMonGenInfo.monGenList[i].Map);
```

RyanTaplin1705:
Ok I would like to think  it's fixed however I have a sheep stood in my map when there is no reference to Sheep in this maps RespawnInfo?

image

[0spawns.txt](https://github.com/Suprcode/mir2/files/2577786/0spawns.txt)

PR #303, opened 2019-05-03
Niggle fixes.

Pete107:
A few niggle fixes.

- Slow now longer persists after the effect has worn off.

- Slow now caps so they don't become inactive due to duration.

- Movement modified (CanRun) on Client when a user exceeds their wear capacity.


Pete107:
> A few niggle fixes.
> 
> * Slow now longer persists after the effect has worn off.
> * Slow now caps so they don't become inactive due to duration.
> * Movement modified (CanRun) on Client when a user exceeds their wear capacity.

Pete107:
Slow no longer persists is what the first point was meant to say.

PR #304, opened 2019-05-08
Pickaxe dura 0 exploit & additional slow fix.

Pete107:
- Prevent Mining from working with Dura 0 pickaxe (additional code to break the pickaxe if you wish to use it just uncomment it)

- Reset the Attack Time upon slow wearing off.


Pete107:
Just a note with this commit, it includes what I see as a fix for Weapon's being damaged when being hit (DamageDura) however it wasn't meant to go over until I got confirmation to push it.

I will revert that change upon request.

PR #305, opened 2019-05-09
add two buttons : reloadAllNPCs and reload Drops

suixin567:
thank you

### DIFF
--- a/Client/AMain.cs
+++ b/Client/AMain.cs
@@ -78,7 +78,7 @@ namespace Launcher
                 }
                 else
                 {
-                    MessageBox.Show("Could not get Patch Information.");
+                    MessageBox.Show(GameLanguage.PatchErr);
                     Completed = true;
                     return;
                 }
@@ -112,7 +112,7 @@ namespace Launcher
         
 
         private void BeginDownload()
-        {
+        {           
             if (DownloadList == null) return;
 
             if (DownloadList.Count == 0)
--- a/Client/CMain.cs
+++ b/Client/CMain.cs
@@ -73,7 +73,7 @@ namespace Client
 
         private void CMain_Load(object sender, EventArgs e)
         {
-
+            this.Text = GameLanguage.GameName;
             try
             {
                 ClientSize = new Size(Settings.ScreenWidth, Settings.ScreenHeight);
@@ -654,7 +654,7 @@ namespace Client
         {
             if (CMain.Time < GameScene.LogTime)
             {
-                GameScene.Scene.ChatDialog.ReceiveChat("Cannot leave game for " + (GameScene.LogTime - CMain.Time) / 1000 + " seconds.", ChatType.System);
+                GameScene.Scene.ChatDialog.ReceiveChat(string.Format("{0}{1}{2}", GameLanguage.CannotLeaveGame, (GameScene.LogTime - CMain.Time) / 1000, GameLanguage.seconds), ChatType.System);
                 e.Cancel = true;
             }
         }
--- a/Client/CMain.cs
+++ b/Client/CMain.cs
@@ -654,7 +654,7 @@ namespace Client
         {
             if (CMain.Time < GameScene.LogTime)
             {
-                GameScene.Scene.ChatDialog.ReceiveChat(string.Format("{0}{1}{2}", GameLanguage.CannotLeaveGame, (GameScene.LogTime - CMain.Time) / 1000, GameLanguage.seconds), ChatType.System);
+                GameScene.Scene.ChatDialog.ReceiveChat(string.Format("{0}{1}{2}", GameLanguage.CannotLeaveGame, (GameScene.LogTime - CMain.Time) / 1000, GameLanguage.Seconds), ChatType.System);
                 e.Cancel = true;
             }
         }
--- a/Client/CMain.cs
+++ b/Client/CMain.cs
@@ -654,7 +654,7 @@ namespace Client
         {
             if (CMain.Time < GameScene.LogTime)
             {
-                GameScene.Scene.ChatDialog.ReceiveChat(string.Format("{0}{1}{2}", GameLanguage.CannotLeaveGame, (GameScene.LogTime - CMain.Time) / 1000, GameLanguage.Seconds), ChatType.System);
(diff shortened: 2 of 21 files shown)


suixin567:
The all language strings is in a single .ini file, convenient for GM of other language

Pete107:
It's a nice starting point for extracting single language.
I personally would allow it to support multiple languages but this would be a lengthy task as there's a lot of hard-coded strings that'd need translating and added.

May even be worth creating an indexed string table.

suixin567:
i think move all language strings in a single .ini file is enough ，diffent language 's GM only need 
translate this file . Support for all languages. I've thought about that before, but I don't think it's necessary.

suixin567:
I wonder if I'm doing the right job？ and  I hope to get a merga.   most commonly used strings is collected now.

suixin567:
@Suprcode

CozarXXVIII:
been thinking this could be an excellent idea, if you could have multiple languages used at same time ( basically changing the language in Settings)

suixin567:
@CozarXXVIII  is a good idea, but it is not necessary.   because In terms of language  hard-decoded strings is only a little part of total .   ths others are so many db tables、all scripts、 drops and so on. For a new language gm , the work of thranslate is unavoidable .   so only independent hard-coded strings is enough . thank you

CozarXXVIII:
@suixin567 While the concept of the idea is good. but the method taken is a bit half arsed. The direction he is currently going is that its moving towards Dephi's String.ini which just moves the server towards another primary single language. For anything like this to be a success and beneficial to all users would be to allow for multiple language strings and allowing language change by a single index either by numeral or language name via Mir2Config.

Suprcode:
I'm in agreement. Whilst i appreciate the client would be better if it could be quickly translated (which you're doing). You also have the issue that many messages are server side, so you have 2 translations to maintain.

I see many of your variable replacements are also not consistant. Some require you to use a colon, others keep the colon hardcoded. All variables should be consistant, and ideally any which use parameters or method calls should use string format to all full edit-ability (and not just part of the string).

Happy for something like this to be added to the crystal source, but not until its in a more complete and polished state.

suixin567:
**I see many of your variable replacements are also not consistant. Some require you to use a colon, others keep the colon hardcoded. All variables should be consistant,**

I don't understand this sentence, could you plaese give me a example code format? I will finish the commonly used strings in two weeks

suixin567:
@CozarXXVIII    i'm a foreigners for you , when i try to play this game  ,i find so many  hard-encoded strings every where ,Some don't even hard to notice, so I just try to make them together .  and... support for all language is not  necessary , bacause is a very very big work !  and some strings even in pictures ,how can we translate it? and a lot of npc scripts can not translate too  and so on .  so what we can do ,is only put the commonly used strings together  and this is enough.

suixin567:
@Suprcode 
Any good ideas？my work is just collect all strings need translate .
Actually ,no matter single or multiple language supported  both need  collect all strings together first.
and Even if my format is wrong, but my work is not wasted
and In fact , My method works well.
I need a merger that would be great. thank you very much. hei~hei~

Suprcode:
> **I see many of your variable replacements are also not consistant. Some require you to use a colon, others keep the colon hardcoded. All variables should be consistant,**
> 
> I don't understand this sentence, could you plaese give me a example code format? I will finish the commonly used strings in two weeks

So for example you have this change

`Hint = GameLanguage.Inventory + " (" + CMain.InputKeys.GetKey(KeybindOptions.Inventory) + ")"`

The language variable is contained just to the first part - restricting how someone would translate the text.

Ideally the language text should be

`GameLanguage.Inventory = "Inventory ({0})"`

And the code should be changed to

`Hint = string.Format(GameLanguage.Inventory, CMain.InputKeys.GetKey(KeybindOptions.Inventory))`

Allowing translation and word inserts at any point in the string.

Unless its consistent through it would be difficult to translate correctly.

suixin567:
thans for you example , and I had try done that , but there is a problem ,some  variables in .ini file will like this : 
GameLanguage.MAC = "MAC: + {0}~{1} (+{2})" : "MAC: + {0}~{1}"
It looks like complicated for a GM .
so I took the middle way: 
a variable still only control  a part of sentence . and the usage  format is consistant now.
How do you like this? if do not like ,I will change it to exactly the same as your suggestion

suixin567:
@Suprcode

Suprcode:
@suixin567 I think having the full string in language is still the better option to allow full flexibility for translation (someone may want to put MAC on the end of the text instead for example).

Its still understandable in the form of "MAC: + {0}~{1} (+{2})" i think. And tbh this will be usually translated by the server owner, not players anyway.

suixin567:
@Suprcode 
The format is the same as you said~ you are right!

suixin567:
@Suprcode  my code is more and more , hope take a merger.

Suprcode:
> @Suprcode my code is more and more , hope take a merger.

Once you're all done with the language strings you're adding let me know and i will review then accept the code

suixin567:
@Suprcode  i am sorry。It's bad for my job.  I can not aways do the work of translating ,  because I have my own  project forked form you , In order for him to operate , need a lot works to do . if you do not give me a merger , how can I push other modules?   I had to make many  branchs , and do repetitive works . and the translating work is not easy to find all of strings , because even some modules I haven't played. and I don't want to play  . so is hard for me to find.  thank you  . I really wan to make a lot of 
contributions , now is get in the way of my work .

Suprcode:
@suixin567 No problem. I will review the code shortly and accept.

suixin567:
@Suprcode  thank you ,  that's great. the next  I will make a lot of
contributions ,  Please trust my speed . I will do my best to make this project a full-fledged for Online products.

Suprcode:
@suixin567 I have reviewed the code and added a few suggestions. Once updated i will accept the code. Thanks

Client/AMain.cs
```diff
@@ -78,7 +78,7 @@ public void Start()
                 }
                 else
```
suixin567: test

suixin567:
@Suprcode  sorry , where is you suggestions? I don't know where it is.

Suprcode:
@suixin567 I would expect you to receive a notification for it.

But you can see this url and scroll down to see each comment

https://github.com/Suprcode/mir2/pull/305/files

suixin567:
@Suprcode  formats is updated , please appect ~ and so I can submit other functions.

PR #309, opened 2019-05-21
Two new features

suixin567:
1、npc script add a new ACT command : OPENBROWSER
aim:
user open a website with browser form npc
usage:
#ACT
OPENBROWSER www.google.com
2、the maps add a attributes : nobackcity . if checked ,players can not use back city scroll
3、add 5 language strings
4、fiexed AMain UI offset error.

### DIFF
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -12880,6 +12880,13 @@ namespace Server.MirObjects
                                 return false;
                             }
                             break;
+                        case 1:
+                            if (CurrentMap.Info.NoEscape)
+                            {
+                                ReceiveChat("You cannot back to city here", ChatType.System);
+                                return false;
+                            }
+                            break;
                         case 2:
                             if (CurrentMap.Info.NoRandom)
                             {
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -12927,7 +12927,7 @@ namespace Server.MirObjects
                         case 1:
                             if (CurrentMap.Info.NoTownTeleport)
                             {
-                                ReceiveChat(GameLanguage.CanNotBackCity, ChatType.System);
+                                ReceiveChat(GameLanguage.NoTownTeleport, ChatType.System);
                                 return false;
                             }
                             break;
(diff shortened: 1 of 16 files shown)


Pete107:
Is a whole class dedicated to finding the default browser necessary?
If I'm not mistaken Process.Start(url) will use the default browser..

Server/MirObjects/PlayerObject.cs
```diff
@@ -12920,21 +12920,28 @@ private bool CanUseItem(UserItem item)
                         case 0:
                             if (CurrentMap.Info.NoEscape)
                             {
-                                ReceiveChat("You cannot use Dungeon Escapes here", ChatType.System);
+                                ReceiveChat(GameLanguage.CanNotDungeon, ChatType.System);
+                                return false;
+                            }
+                            break;
+                        case 1:
+                            if (CurrentMap.Info.NoEscape)
```
Suprcode: this looks like a bug. It should be checking "BackCity"

Suprcode:
Please can all references to "NoBackCity" be replaced with "NoTownTeleport" to stay consistent with its use.

Suprcode:
> Is a whole class dedicated to finding the default browser necessary?
> If I'm not mistaken Process.Start(url) will use the default browser..

Yes that would open your native IE browser - however there is functionality in the class to look for chrome first and open chromes process with a default url

suixin567:
@Suprcode I am  sorry  the bug is fixed.  and the name is changed too. 
and the feature about OpenbrowserHelper  class will not work right all the time ,because some  rogue software. but I think is enough now. 
so I hope a merger . And ,  I have written a another new feature make game server support Http service ,because  I need my website allow user login with the crystal game account, and so on... because there have no  DB like mysql 、sqlLite ... so , add a http service is convenient . what do you think?  if think so , I want to make a push .

Suprcode:
> @Suprcode I am sorry the bug is fixed. and the name is changed too.
> and the feature about OpenbrowserHelper class will not work right all the time ,because some rogue software. but I think is enough now.
> so I hope a merger . And , I have written a another new feature make game server support Http service ,because I need my website allow user login with the crystal game account, and so on... because there have no DB like mysql 、sqlLite ... so , add a http service is convenient . what do you think? if think so , I want to make a push .

Sounds fine. Please add as a separate library if it requires several dependencies or classes though and ensure it is able to be turned off / configurable through an ini file / menu form.

PR #312, opened 2019-05-24
add http service

suixin567:
@Suprcode 
aim:
 Let the web server communicate with the game server
route:
1、/regist   regist a new account by way of website form.
2、/login  login website with game account
3、/addNameList  add a nameList by way of website
and there is a trusted ipaddress can be config , for security.

### DIFF
--- a/Server/MirEnvir/Envir.cs
+++ b/Server/MirEnvir/Envir.cs
@@ -527,8 +527,11 @@ namespace Server.MirEnvir
                 }
 
                 StartNetwork();
-                http = new HttpServer();
-                http.Start();
+                if (Settings.StartHTTPService)
+                {
+                    http = new HttpServer();
+                    http.Start();
+                }               
                 try
                 {
                     while (Running)
@@ -1842,8 +1845,7 @@ namespace Server.MirEnvir
                     MobThreading[i].Interrupt();
                 }
             }
-            http.Stop();
-
+                http.Stop();        
 
                 while (_thread != null)
                     Thread.Sleep(1);
--- a/Server/MirEnvir/Envir.cs
+++ b/Server/MirEnvir/Envir.cs
@@ -2169,6 +2169,56 @@ namespace Server.MirEnvir
                 c.Enqueue(new ServerPackets.NewAccount {Result = 8});
             }
         }
+
+        public int HTTPNewAccount(ClientPackets.NewAccount p, string ip)
+        {
+            if (!Settings.AllowNewAccount)
+            {                
+                return 0;
+            }
+
+            if (!AccountIDReg.IsMatch(p.AccountID))
+            {
+                return 1;
+            }
+
+            if (!PasswordReg.IsMatch(p.Password))
+            {
+                return 2;
+            }
+            if (!string.IsNullOrWhiteSpace(p.EMailAddress) && !EMailReg.IsMatch(p.EMailAddress) ||
+                p.EMailAddress.Length > 50)
+            {
+                return 3;
+            }
+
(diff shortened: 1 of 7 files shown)


Server/Settings.cs
```diff
@@ -59,6 +59,9 @@ internal static class Settings
                              RelogDelay = 50,
                              MaxIP = 5;
 
+        //HTTP
+        public static string HTTPIPAddress = "http://127.0.0.1:5679/";
+        public static string HTTPTrustedIPAddress = "127.0.0.1:5679";
```
Suprcode: what is this HTTPTrustedIPAddress used for? It has no implementation from what i can see.

Server/MirEnvir/Envir.cs
```diff
@@ -526,7 +527,8 @@ private void WorkLoop()
                 }
 
                 StartNetwork();
-
+                http = new HttpServer();
```
Suprcode: please put this under a config setting to toggle starting the httpserver

Server/Utils/HttpService.cs
```diff
@@ -0,0 +1,89 @@
...
+            if (listener != null && listener.IsListening)
+            {
+                listener.Stop();
+            }           
+        }
+
+        public abstract void OnGetRequest(HttpListenerRequest request, HttpListenerResponse response);
+        public abstract void OnPostRequest(HttpListenerRequest request, HttpListenerResponse response);
+
+        public void WriteRresponse(HttpListenerResponse response, string responseString)
```
Suprcode: how come its WriteRresponse, and not WriteResponse?

Suprcode:
thanks. Just a question on the WriteRresponse naming. And still need a config setting to not start the http service.

Many people won't want this enabled so should be off by default.

suixin567:
@Suprcode  
Toggle have been added, and default is off. 
the next my plan is :
1、support chanese character name
2、make game class can be off , because the assassins and archers are unfamiliar for most of us.
And I have a wondering about : how to make players do not need install a DirectX SDK ? if do not install ,there will a black screen... thanks very much.

Suprcode:
You can already turn off archers and assassin in the code. This is in the config menu.

No need for built in DirectX. The 4 dll files needed are already supplied with the client. No need for sdk

suixin567:
@Suprcode  sorry I do not understand , is there a toggle in config menu to off class ?  if so ,I don't have to do it again.
and the 4 dll files do not work ,  I tried it on two computers, both are black with that 4 dll files ~~~

Suprcode:
If you're still having a black screen after the 4 dll's - then yes you do need to install it.

I expect its different depending on your hardware. I don't have to install the SDK on mine. The 4 dll's are the only extra ones i need.

In config menu -> Permissions

You can disable assassin and archer class already

Server/MirEnvir/Envir.cs
```diff
@@ -526,7 +527,11 @@ private void WorkLoop()
                 }
 
                 StartNetwork();
-
+                if (Settings.StartHTTPService)
+                {
+                    http = new HttpServer();
```
Suprcode: this instance is no longer required as you're creating it in the constructor

suixin567:
@Suprcode sorry!  instance problem haved fixed ! I was careless

Suprcode:
> @Suprcode sorry! instance problem haved fixed ! I was careless

I'm not able to see this change yet? Still creates 2 instances.

suixin567:
@Suprcode  the one is removed from constructor.  the instance is only create when setting is on.

suixin567:
forget to push.......

PR #313, opened 2019-05-25
About Monster Info Form

suixin567:
Add monster fileName display on MonsterInfoForm


suixin567:
@Suprcode  
There are too many monsters, and also make client file size too big , so when some one want to delete monsters  this feature will helpful.

PR #315, opened 2019-05-26
Character Start Info,

CozarXXVIII:
Character Start info added to Config.tab -> Server -> Optional

Start Level & Start Gold

/daneo1989


Suprcode:
I don't see any requirement in moving the projects to new folders. This will only disrupt existing users.

I'm also reluctant to accept built in start level and gold through the config. Both of these can be achieved in a better way via the 00default npc - and having a higher than zero start gold built in to every user without account checks simply opens up the server to exploitation.

PR #319, opened 2019-05-28
Fix and small refactor.

Pete107:
Fixed Fatal sword from leveling before knowing if the hit landed or not.

Extracted the Armour calculation in the Attacked methods into MapObject, can be overridden.
Extracted applying negative effects in the Attacked methods (player vs object)

Should make the Attacked methods look a little cleaner.

### DIFF
--- a/Server/MirObjects/MapObject.cs
+++ b/Server/MirObjects/MapObject.cs
@@ -399,6 +399,71 @@ namespace Server.MirObjects
         public abstract int Attacked(PlayerObject attacker, int damage, DefenceType type = DefenceType.ACAgility, bool damageWeapon = true);
         public abstract int Attacked(MonsterObject attacker, int damage, DefenceType type = DefenceType.ACAgility);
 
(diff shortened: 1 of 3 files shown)


Server/MirObjects/MapObject.cs
```diff
@@ -399,6 +399,71 @@ public virtual void BroadcastInfo()
         public abstract int Attacked(PlayerObject attacker, int damage, DefenceType type = DefenceType.ACAgility, bool damageWeapon = true);
         public abstract int Attacked(MonsterObject attacker, int damage, DefenceType type = DefenceType.ACAgility);
 
+        public virtual void GetArmour(DefenceType type, MapObject attacker, out bool hit, ref int armour)
```
Suprcode: Since this method is a GetArmour method - it should really be returning the armour value set - instead of having it as a ref parameter

Pete107: All should be good now.

PR #320, opened 2019-05-28
Little changes

suixin567:
1、Change client c# version from 4 to 6 , because when in release mode will appear errors.
2、Add 5 language strings.


Client/MirScenes/GameScene.cs
```diff
@@ -7756,7 +7756,7 @@ public MirControl OverlapInfoLabel(UserItem item, bool Inspect = false)
                     Location = new Point(4, ItemLabel.DisplayRectangle.Bottom),
                     OutLine = true,
                     Parent = ItemLabel,
-                    Text = string.Format(GameLanguage.MaxCombine, realItem.StackSize, GameLanguage.SplitStack)
+                    Text = string.Format(GameLanguage.MaxCombine, realItem.StackSize, "\n")
```
Suprcode: remove the \n newline - this should go within the string .MaxCombine so its optional to have a new line / remove it if preferred

Common.cs
```diff
@@ -6464,8 +6464,7 @@ public class GameLanguage
                          Password = "Password",
 
                          ShuttingDown = "Disconnected: Server is shutting down.",
-                         MaxCombine = "Max Combine Count : {0}{1}",
-                         SplitStack = "Shift + Left click to split the stack",
+                         MaxCombine = "Max Combine Count : {0}{1}Shift + Left click to split the stack",
```
Suprcode: remove the second parameter {1} and replace with \n

suixin567:
@Suprcode  
1、Sorry, "\n" can not move in to variable because a “\n” will be no effect when be loaded from .ini file. 
2、Continue add 9 language strings
3、I find that the file MonsterFile in Server Project is in error now, about at line 2100 . have some one  do some damages?

Suprcode:
ok no problem about \n.

There are no build issues for me. Make sure you have merged correctly (and pulled latest code)

Just 1 more question before i can merge this

PR #323, opened 2019-06-01
Add DX libs to Components forlder

suixin567:
1、Add DriectX lib files to mir/Components forlder
2、8 languages strings


Suprcode:
3rd party library files should never be checked in with source code.

You can get these through the client autopatcher - which is a requirement for the rest of the client anyway, so there is no need to have these here.

Please remove the library file additions and i can accept the remaining pull request

suixin567:
@Suprcode  
1、removed dx dll files
2、continue add more than forty strings.

Client/Client.csproj
```diff
@@ -97,23 +97,19 @@
...
-      <HintPath>..\..\Client\DirectX\Microsoft.DirectX.Direct3D.dll</HintPath>
-      <Private>False</Private>
+      <HintPath>..\..\..\..\..\Windows\Microsoft.NET\DirectX for Managed Code\1.0.2902.0\Microsoft.DirectX.Direct3D.dll</HintPath>
     </Reference>
-    <Reference Include="Microsoft.DirectX.Direct3DX, Version=1.0.2909.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
+    <Reference Include="Microsoft.DirectX.Direct3DX, Version=1.0.2902.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
       <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\..\Client\DirectX\Microsoft.DirectX.Direct3DX.dll</HintPath>
-      <Private>False</Private>
+      <HintPath>..\..\..\..\..\Windows\Microsoft.NET\DirectX for Managed Code\1.0.2902.0\Microsoft.DirectX.Direct3DX.dll</HintPath>
```
Suprcode: Why this change in hint path? wouldn't the local client directory be more suited - not all people (including myself) will have the directx stored in that path, so would have to manually change the path.

PR #330, opened 2019-06-27
Potential exploit fixed

Pete107:
Fixed a potential exploit on retrieving gold from guild storage where the client was the only one doing the validation!

### DIFF
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -15875,6 +15875,11 @@ namespace Server.MirObjects
             }
             else
             {
+                if (MyGuildRank != MyGuild.Ranks[0])
+                {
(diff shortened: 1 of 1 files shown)


Suprcode:
There is already a check below yours which does this

`if (MyGuildRank.Index != 0)
                {
                    ReceiveChat("Insufficient rank.", ChatType.System);
                    return;
                }`

Pete107:
> There is already a check below yours which does this
> 
> `if (MyGuildRank.Index != 0) { ReceiveChat("Insufficient rank.", ChatType.System); return; }`

Yeah I seen that a little further down just now, my bad!

Any reason why this isn't attached to any RankOptions such as retrieve items?

Suprcode:
i expected it would be a rank option tbh.

PR #337, opened 2019-10-01
Fixed two bugs 

suixin567:
Fixed httpService and fixed a big bug in MapObject.cs


suixin567:
aaaaaaaaaaaaaa

Pete107:
The armor bug was fixed in my last commit you can also see in the current code it is fine.

https://github.com/Suprcode/mir2/blob/master/Server/MirObjects/MapObject.cs#L432

Issue #339, opened 2019-10-18
Character names cause Reporting crash


louisnorthmore:
If you name it as below, the report folder will not be created, causing a server crash.

* Character name

CON
AUX
PRN
NUL
COM0 ~ COM9
LPT0 ~ LPT9

https://www.lomcn.org/forum/threads/folders-unable-to-create-in-windows-server-crash.103648/#post-1140940
https://www.lomcn.org/forum/threads/can-i-have-help-please.103656

Crash:
`System.IO.DirectoryNotFoundException: Could not find a part of the path. at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath) at System.IO.Directory.InternalCreateDirectory(String fullPath, String path, Object dirSecurityObj, Boolean checkHost) at System.IO.Directory.InternalCreateDirectoryHelper(String path, Boolean checkHost) at System.IO.Directory.CreateDirectory(String path) at Server.MirEnvir.Reporting..ctor(PlayerObject player) in D:\WarCraft 2\SERVER\WarCraft\Server\MirEnvir\Reporting.cs:line 44 at Server.MirObjects.PlayerObject..ctor(CharacterInfo info, MirConnection connection) in D:\WarCraft 2\SERVER\WarCraft\Server\MirObjects\PlayerObject.cs:line 413 at Server.MirNetwork.MirConnection.StartGame(StartGame p) in D:\WarCraft 2\SERVER\WarCraft\Server\MirNetwork\MirConnection.cs:line 1101 at Server.MirNetwork.MirConnection.ProcessPacket(Packet p) in D:\WarCraft 2\SERVER\WarCraft\Server\MirNetwork\MirConnection.cs:line 250 at Server.MirNetwork.MirConnection.Process() in D:\WarCraft 2\SERVER\WarCraft\Server\MirNetwork\MirConnection.cs:line 187 at Server.MirEnvir.Envir.WorkLoop() in D:\WarCraft 2\SERVER\WarCraft\Server\MirEnvir\Envir.cs:line 831 at line 0`

PR #340, opened 2019-10-18
Character Names crashing server

louisnorthmore:
Character names outlined in #339 crashing server. 

This places a try/catch block around the `File.Create` and adds the exception to the error log in the normal way.


louisnorthmore:
Added to also append `_player` to player report folder name. This ensures the folder name will not be a protected namespace. 
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

PR #341, opened 2019-10-19
NoGroup - Map attribute.

daneo1989:
NoGroup map attribute

### DIFF
--- a/Server/MirDatabase/MapInfo.cs
+++ b/Server/MirDatabase/MapInfo.cs
@@ -17,7 +17,7 @@ namespace Server.MirDatabase
 
         public bool NoTeleport, NoReconnect, NoRandom, NoEscape, NoRecall, NoDrug, NoPosition, NoFight,
             NoThrowItem, NoDropPlayer, NoDropMonster, NoNames, NoMount, NeedBridle, Fight, NeedHole, Fire, Lightning, 
-            NoTownTeleport, NoReincarnation;
+            NoTownTeleport, NoReincarnation, NoGroup = false;
 
         public string NoReconnectMap = string.Empty;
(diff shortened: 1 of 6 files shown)


Pete107:
May I suggest toggling the AllowGroup to false rather than back to true.

Suprcode:
Why do we need to change the players own AllowGroup setting at all? There should be logic which just prevents them from adding group members (which it looks like you've done). Then just leave any existing group they're in.

Also prevent them from joining a group when they're on the map.

daneo1989:
I looked into the AllowGroup as suggested. (mostly around SwitchGroup in PlayerObject), 

might be worth also making it so the ClientSide disables the AllowGroup button but the fundamental object in disabling the Invites, Adding, Removing on changing maps is there.

Pete107:
> I looked into the AllowGroup as suggested. (mostly around SwitchGroup in PlayerObject),
> 
> might be worth also making it so the ClientSide disables the AllowGroup button but the fundamental object in disabling the Invites, Adding, Removing on changing maps is there.

Yeah, that's what I meant, just disabling the toggle for the player in general on maps with the attribute by setting the AllowGroup to false, it's just a visual effect is all.

Suprcode:
> I looked into the AllowGroup as suggested. (mostly around SwitchGroup in PlayerObject),
> 
> might be worth also making it so the ClientSide disables the AllowGroup button but the fundamental object in disabling the Invites, Adding, Removing on changing maps is there.

You're changing the players setting though. Theres no need. Just prevent the logic from happening in adding members / joining groups.

Suprcode:
Due to another merge this can no longer be commited. Please correct the change suggested and resubmit the pull request.

PR #342, opened 2019-10-20
Server library for .NET Core

MirageCode:
**There shouldn't be any functional change in this pull request.** The purpose of it is to isolate (not remove, yet) the use of `System.Windows.Forms` in the server so that most of the code is more portable. This makes it easier to write tools in .NET Core which reuse this code, and perhaps ultimately write a replacement for `Server.MirForms`.

Here are the key points of what I did:

- Factor `Server.MirForms` out of `Server`
- Move the entrypoint `Program.cs` and `Server.csproj` into `Server.MirForms`
- Factor some of the globals in `SMain` into singletons (in `Envir` and `MessageQueue`)
- Move a small amount of code from `Server.MirDatabase` (`AccountInfo` and `CharacterInfo`) to `Server.MirForms`
- Make `Server` a classlib targetting .NET Standard 2.0 and .NET Core 2.1 so that it builds on both .NET Core and Mono.
- Update `Server.MirForms` from .NET Framework 4.5 to 4.6.1 to make it compatible with .NET Standard 2.0 so that it can reference the `Server` classlib and it still builds on Mono. It doesn't build on .NET Core but now it is isolated.

I've seen some of the work done by @Pete107 which is great and helped a lot when checking over this. I've tried to maintain some consistency with his work to make it easier for others to understand. But what this does which his doesn't is actually get the old `Server.MirForms` interface to work with the server library. It doesn't add a new .NET Core entrypoint, like his does, but one can easily be built on top of this and merged at the point where it is fully functional. And, as this doesn't duplicate code, it makes it easy to pull any changes downstream while working on separate entrypoints.


louisnorthmore:
Loving the idea behind this. However people won't know how to get this running if they have an existing solution running. A git fetch,pull results in a mast branch that cannot be built.

Do we need some kind of documentation for this and how is the best way to do that?

MirageCode:
@louisnorthmore `msbuild -restore Server.MirForms/` works fine for me on Mono. The client has got the same problem with icon resources on Mono but, if I remove those, `msbuild -restore` on its own builds the whole thing. `dotnet msbuild -restore Server/` builds the server library on .NET Core.

Admittedly, I'm running this on Linux and haven't actually got a Windows development environment to test it on. What errors do you get when trying to build?

louisnorthmore:
@MirageCode It worked once I had built the new server project and had the .dll. 

I'm guessing most people would just double-click the .sln file to load visual studio.

louisnorthmore:
Great work btw! 👍

Suprcode:
@MirageCode I've just pulled this to confirm what @louisnorthmore said. The solution currently doesn't have the the newly separate Server.Library project added - so fails to build until you manually add the project.

Please can you commit the updated .sln to fix this

MirageCode:
@louisnorthmore Thanks.

The `Server.csproj` in `Server.MirForms` references the `Server.Library.csproj`, so the dll gets built automatically when I build the exe. And `Server.csproj` is referenced in the .sln, which is how everything builds when I run `msbuild -restore`.

I would've thought it would just work on Windows too. Though, as said, I haven't tried it. If it doesn't, perhaps `Server.Library.csproj` needs to be referenced in the .sln?

MirageCode:
@Suprcode Thanks. Just done on the same branch. Can you confirm this resolves it? If so, do you need me to create a new pull request?

PR #343, opened 2019-10-22
Add Server.Library.csproj to .sln file

MirageCode:
This should fix building in Visual Studio on Windows, though I haven't tested it.


louisnorthmore:
Tested. Looks good here.

PR #349, opened 2019-12-12
Fix Map Lightning & Lava

Pete107:
Map Lightning and Lava will now deal damage to player(s) previously removed by mistake.


donz187:
I

PR #362, opened 2020-08-10
Archer fix  ( and npc play sound action)

mrgreaper:
This is a change to the rangeattack in playerobject so that it uses either mc (the default of the core files) or dc(what the bows are set up to deal damage with in the commonly used databases.
It works out which is the higher stat and then uses that stat.
so if you have a bow with dc on it and your character is a dc based character you will do DC min / max
but if you are a mc based character with mc stated bows you will do mc min/max

This should allow for expanding the ranged characters later and making for a more versatile system (in my opinion)

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -1708,6 +1708,9 @@ namespace Client.MirScenes
                 case (short)ServerPacketIds.OpenBrowser:                  
                     OpenBrowser((S.OpenBrowser)p);
                     break;
+                case (short)ServerPacketIds.playSound:
+                    playThesound((S.playSound)p);
+                    break;
                 default:
                     base.ProcessPacket(p);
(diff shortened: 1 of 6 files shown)


mrgreaper:
Did try to do this as a seperate pull request but they have added to this pull:
npc action "play sound" 
and playerobject function for playing sound. (used by the npc action and handy for a lot of things)
as per my tutorial https://www.lomcn.org/forum/threads/how-to-make-an-npc-make-sound.104734/

Suprcode:
This has been added in the commit below.

https://github.com/Suprcode/mir2/commit/7c533b60a276c9e27cb6ebb61b240c5c4461d943

Please remember to use correct casing conventions on commits.

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions?redirectedfrom=MSDN

PR #373, opened 2020-09-03
BlackScreen/ ScreenPosition on first login - Fix

meacher0:
(Possible) Fix for BlackScreen and/or incorrect screen positioning when entering game from Char Select whilst in fullscreen.


meacher0:
https://www.lomcn.org/forum/threads/minimum-resolution-increased.104771/page-2

PR #376, opened 2020-11-13
Add Alt+Click to sell support

cgluckg:
This should be part of the core files, so players are able to sell quicker.


Suprcode:
Always thought that was the whole point of the "hold" button. But happy to commit this anyway

cgluckg:
Using "Hold" you still need to drag the items to the "sell window".
With this, you just keep Alt key pressed and click on the item you want to sell.
Makes it especially easy if you have a bag full of stuff to sell.

Anyway... thanks for merging!

PR #378, opened 2020-11-23
Converted to SharpDX

Essensoft-Roc:
1. Get rid of SlimDX dependency vc runtime(Microsoft Visual C++ 2010 Redistributable x64).
2. Preparing for migration to .NET 5.


Suprcode:
I don't see any benefits gained from converting the project to SharpDX (of which has only recently been upgraded to use SlimDX).

Essensoft-Roc:
I think SharpDX may be better. 
If upgraded to .NET Core or .NET 5.

Suprcode:
Better in what way?

SlimDX should be fully supported on .NET 5 - And you can't convert the client to use Core because of other dependencies on Win Forms.

You're not having any performance gain, and both support 64 bit.

Looking up reading material SharpDx requires more manual memory handling, so likely will pose issues if not managed correctly (unlike SlimDx which handles it for you).

I'm very relucant to merge such a major change unless there are noticable improvements gained from moving.

jxsr099:
image
Please solve it

roc916:
> image
> Please solve it

is closed

PR #380, opened 2020-11-24
Map Coords shown below Big Map Dialog

cgluckg:
When mouse is over BigMapDialog, just below the map the actual map coordinates are shown as the mouse is moved.


cgluckg:
Missed some code to show coords right below map, depending on the map size.

PR #381, opened 2020-11-24
Updated BigMapDialog with map coords

cgluckg:
Added map coordinates on mouse move over BigMapDialog.

### DIFF
--- a/Client/MirScenes/Dialogs/MainDialogs.cs
+++ b/Client/MirScenes/Dialogs/MainDialogs.cs
@@ -4840,16 +4840,11 @@ namespace Client.MirScenes.Dialogs
     public sealed class BigMapDialog : MirControl
     {
     	public static Point BigMapMouseLocation;
-        public MirLabel BigMapLocationLabel;
 	
-	Rectangle viewRect;
-        float scaleX;
-        float scaleY;
+	float ScaleX;
+        float ScaleY;
 	
         int BigMap_MouseCoordsProcessing_OffsetX, BigMap_MouseCoordsProcessing_OffsetY;
-    
-    	int BigMapCoordsLabelWidth = 60;
-        int BigMapCoordsLabelHeight = 20;
             
         public BigMapDialog()
         {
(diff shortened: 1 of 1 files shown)


SpartaMir:
Tested this, no cords show ingame

cgluckg:
> Tested this, no cords show ingame

I missed NotControl = false; This is by default set to true.

KingdomMir2:
Could you add it so the cord follow mouse pointer on big map?

SpartaMir:
yes following mouse pointer would be good, hard to see the cords on lighter maps, can confirm it works great now.

cgluckg:
I will give it a try tomorrow to try and make it follow the mouse. I will re-post when I have a solution. Thanks for the idea!

cgluckg:
Managed to get this done to have it move with the pointer.
However, it looks laggy and it is dependent on the cursor pixels
What do you guys think?

Suprcode:
Ideally you should be creating something like the magic hover labels that follow the mouse around - they're much more smooth than constantly updating the label's position in draw.

Also will have been cleaner to just make ViewRect and ScaleX/ScaleY public, instead of making a second lot of variables to store this information

cgluckg:
I tried looking where that label is created but couldn't find the exact spot.

Regarding the public definition of ViewRect and ScaleX/ScaleY, this would be possible, as far as I see, but only for ScaleX and ScaleY.
ViewRect.Location is changed with
      viewRect.Location = new Point(
                (int)(scaleX * MapObject.User.CurrentLocation.X) - viewRect.Width / 2,
                (int)(scaleY * MapObject.User.CurrentLocation.Y) - viewRect.Height / 2);
and I needed the info before this change to be able to calculate the coords.

Please correct me if I'm wrong.
Also, can you give me a hint as to where to start looking for that hover label?

Thanks!

cgluckg:
> I tried looking where that label is created but couldn't find the exact spot.
> 
> Regarding the public definition of ViewRect and ScaleX/ScaleY, this would be possible, as far as I see, but only for ScaleX and ScaleY.
> ViewRect.Location is changed with
> viewRect.Location = new Point(
> (int)(scaleX * MapObject.User.CurrentLocation.X) - viewRect.Width / 2,
> (int)(scaleY * MapObject.User.CurrentLocation.Y) - viewRect.Height / 2);
> and I needed the info before this change to be able to calculate the coords.
> 
> Please correct me if I'm wrong.
> Also, can you give me a hint as to where to start looking for that hover label?
> 
> Thanks!

Managed to get something working, so ignore my previous comment.
The only part that still applies is that I need the offsets, so along defining ScaleX/ScaleY and viewRect as public, I left BigMap_MouseCoordsProcessing_OffsetX and BigMap_MouseCoordsProcessing_OffsetY because I need the offset before the actual location of the map is calculated (at lease this is how I see it).

The label is now more smooth moving together with the mouse, but I have the issue that it's disappearing every 2 moves.
So, I move the mouse 1 pixel, the coords appear, but the next pixel move they disappear and so on.

I will look more into this later today or tomorrow, but maybe you guys have a hint on it.

Suprcode:
The "HintLabel" is already setup to do this on anything you set a hint for.

Since the big map is a control itself, just set the Hint variable to the coords and it should just work.

Also you've made ScaleX/Y public, but you're still declaring the old variables.

int MouseCoordsOnBigMap_MapValue_X = (int)((BigMapDialog.BigMapMouseLocation.X - BigMap_MouseCoordsProcessing_OffsetX) / scaleX);
            int MouseCoordsOnBigMap_MapValue_Y = (int)((BigMapDialog.BigMapMouseLocation.Y - BigMap_MouseCoordsProcessing_OffsetY) / scaleY);

can be changed to just

int MouseCoordsOnBigMap_MapValue_X = (int)((BigMapMouseLocation.X - viewRect.X) / scaleX);
int MouseCoordsOnBigMap_MapValue_Y = (int)((BigMapMouseLocation.Y - viewRect.Y) / scaleY);


Public variables should start with Upper case letters, local variables should start with a lower case letter

cgluckg:
OK, I did the modifications.
Still a bit laggy, but I don't think there's anything else I could do to overcome this.

I tried what you said, to use viewRect.X and viewRect.Y instead of BigMap_MouseCoordsProcessing_OffsetX and BigMap_MouseCoordsProcessing_OffsetY and the coords are not correct like that.
I need to take the initial offset, right after the map is processed and this can only be done via BigMap_MouseCoordsProcessing_OffsetX and BigMap_MouseCoordsProcessing_OffsetY.
Given the above statement, viewRect was not required as public anymore, so I moved it back there it initially was.

In case I missed something, please let me know.

Suprcode:
Just one more thing, you do not need to have 2 events

MouseMove += UpdateBigMapCoordinates;
            MouseMove += (o, e) => BigMapMouseLocation = e.Location;

You can just access e.Location from inside the UpdateBigMapCoordinates method.

This may be why its laggy

cgluckg:
> Just one more thing, you do not need to have 2 events
> 
> MouseMove += UpdateBigMapCoordinates;
> MouseMove += (o, e) => BigMapMouseLocation = e.Location;
> 
> You can just access e.Location from inside the UpdateBigMapCoordinates method.
> 
> This may be why its laggy

You are right!
Changed the code.
Still laggy, but I guess it's better than not having this feature at all.

PR #399, opened 2021-03-29
Made Light sizes bigger due to cut off.

Pete107:
Due to the small size of some Lights Skills such as Mass Healing & Fire Bang were having their Lights cut off causing the Bounds of the Rectangle being visible rather than blending.


Suprcode:
Instead of constantly needing to expand the light source i've just emptied out the limiting rectangle to allow any size it wants. This is how its been done for map lighting and player lighting too.

Pete107:
I hadn't actually noticed that, thanks.

PR #407, opened 2021-04-10
Mob AI 4:

DrowninGCode:
- White Mammoth - Added AI and Spell Animation


DrowninGCode:
VS is being a pain in the ass with commits, not sure what i'm doing wrong (keeping updating everything after a pull request is commited, but it still wants to pull previous commit with it) so i've reset everything and just sent one AI as a test.

PR #417, opened 2021-04-15
Mob AI 11:

DrowninGCode:
### **Added:**
* TucsonEgg AI
* TucsonEgg2 AI (This spawns the TucsonGeneral upon death)

### **Removed:**
* TucsonMage Spell animation from Attack1 (animation is for Attack2 only)

### **Amended:**
* SeedingsGeneral - 285.lib - Removed blank images from DrawEffects frames in lib file.
* TucsonMage - Increased random attack from 3 to 5.
* TucsonWarrior - Fixed Attack1 target attacked and added random attack for attack2 for when mob is in melee range of player.
* Armadillo - 290.lib - Removed blank images from Attack1 frames in lib file.

### DIFF
--- a/Client/MirObjects/MonsterObject.cs
+++ b/Client/MirObjects/MonsterObject.cs
@@ -633,9 +633,6 @@ namespace Client.MirObjects
                             case Monster.SeedingsGeneral:
                                 Effects.Add(new Effect(Libraries.Monsters[(ushort)Monster.SeedingsGeneral], 536 + (int)Direction * 4, 4, Frame.Count * Frame.Interval, this));
                                 break;
-                            case Monster.TucsonMage:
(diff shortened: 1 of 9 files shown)


KingdomMir2:
On the TucsonEgg's would it better to have hits to kill , instead of hp based hp?

DrowninGCode:
> On the TucsonEgg's would it better to have hits to kill , instead of hp based hp?

I can change it to hit-based, that's no problem - i'll make amendments to the AI and push it with my next commit :)

KingdomMir2:
it may make it more feeling your trying to crack a egg :)

DrowninGCode:
Haha, yeah plus I have set up TucsonEgg1 which will spawn the TucsonGeneral, so rather than people one-shotting all the eggs on a map, they will have to attack multiple times and wait for the egg to pop to see if they spawned the boss ^^

PR #452, opened 2021-07-05
Increase the light intensity and modify skills (according to the official server setting)

cjlaaa:
Increase the light intensity (refer to the official version)

### DIFF
--- a/Client/MirGraphics/DXManager.cs
+++ b/Client/MirGraphics/DXManager.cs
@@ -191,10 +191,10 @@ namespace Client.MirGraphics
                             using (PathGradientBrush brush = new PathGradientBrush(path))
                             {
                                 Color[] blendColours = { Color.White,
+                                                     Color.FromArgb(255,200,200,200),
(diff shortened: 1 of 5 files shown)


cjlaaa:
Skill modification (according to the official server setting)
1. The skill PoisonCloud modifies the green poison damage to the average value of the MaxSC and MinSC.
2. The skill Plague modified to single damage, and the mana cost is modified to the sum of the MaxSC and MinSC, and it causes the same damage to the target.
3. The skill ProtectionField increases the MinAC.
4. The skill Rage increases the MinDC and shortens the duration.
5. The skill SoulShield/BlessedArmour/UltimateEnhancer extension duration (according to official experience, not the exact official setting).
6. The skill Fencing does not increase defense.
7. The skill Slaying adds 1 point of accuracy per level, and each level of 0-3 increases 5/6/7/8 points of DC.
8. The skill FatalSword does not increase accuracy.
9. The skill SpiritSword do not increase the DC, and the accuracy of 1-3 levels are increased by 3/5/8 points.

PR #456, opened 2021-07-11
Some modification refer to official settings

cjlaaa:
1. Transform type props can provide Fast Run. (refer to official settings)
2. Pointing the mouse to the monster body will display the name of the monster. (Refer to the official settings)
3. Add the potion of drop buff according to the method of exp buff.

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -6228,6 +6228,10 @@ namespace Client.MirScenes
                 {
                     text = string.Format("Exp + {0}% ", minValue + addValue);
                 }
+                else if (realItem.Type == ItemType.Potion && realItem.Shape == 6)
+                {
+                    text = string.Format("Drop + {0}% ", minValue + addValue);
+                }
                 else
                 {
                     text = string.Format(minValue + addValue > 0 ? GameLanguage.Luck: "Curse + {0}", Math.Abs(minValue + addValue));
@@ -9197,7 +9201,17 @@ namespace Client.MirScenes
             if (MapObject.MouseObject != null && !(MapObject.MouseObject is ItemObject))
                 MapObject.MouseObject.DrawName();
 
-            int offSet = 0;
+            int offSet = 0; 
+            
+            for (int i = 0; i < Objects.Count; i++)
+            {
+                MonsterObject ob = Objects[i] as MonsterObject;
+                if (ob == null) continue;
+
+                if (!ob.MouseOver(MouseLocation)) continue;
+                ob.DrawName();
+            }
+            
             for (int i = 0; i < Objects.Count; i++)
             {
                 ItemObject ob = Objects[i] as ItemObject;
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -6228,7 +6228,7 @@ namespace Client.MirScenes
                 {
                     text = string.Format("Exp + {0}% ", minValue + addValue);
                 }
-                else if (realItem.Type == ItemType.Potion && realItem.Shape == 6)
+                else if (realItem.Type == ItemType.Potion && realItem.Shape == 5)
                 {
                     text = string.Format("Drop + {0}% ", minValue + addValue);
                 }
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -9203,15 +9203,18 @@ namespace Client.MirScenes
 
             int offSet = 0; 
             
-            for (int i = 0; i < Objects.Count; i++)
+            if (Settings.DisplayBodyName)
             {
-                MonsterObject ob = Objects[i] as MonsterObject;
-                if (ob == null) continue;
+                for (int i = 0; i < Objects.Count; i++)
+                {
+                    MonsterObject ob = Objects[i] as MonsterObject;
+                    if (ob == null) continue;
 
-                if (!ob.MouseOver(MouseLocation)) continue;
-                ob.DrawName();
+                    if (!ob.MouseOver(MouseLocation)) continue;
+                    ob.DrawName();
+                }
             }
-            
+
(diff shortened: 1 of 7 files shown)


Server/MirObjects/PlayerObject.cs
```diff
@@ -11291,6 +11292,12 @@ public void UseItem(ulong id)
                                 AddBuff(BuffType.Exp, this, Settings.Minute * time, new Stats { [Stat.ExpRatePercent] = item.GetTotal(Stat.Luck) });
                             }
                             break;
+                        case 6: //Drop
```
Suprcode, line 11295: Why have you used 6? 5 is not used currently.

Client/MirScenes/GameScene.cs
```diff
@@ -9197,7 +9201,17 @@ protected override void CreateTexture()
             if (MapObject.MouseObject != null && !(MapObject.MouseObject is ItemObject))
                 MapObject.MouseObject.DrawName();
 
-            int offSet = 0;
+            int offSet = 0; 
+            
+            for (int i = 0; i < Objects.Count; i++)
```
Suprcode, line 9206: Please add this under a client setting which is defaulted to OFF. Not everyone may want this feature.

Suprcode (changes requested):

cjlaaa:
Some modifications
1. Add a switch to display the name of the monster's body.
2. The damage value of the poison cloud skill has been modified correctly. (Refer to the official settings).
3. Add broadcast of poison damage and healing value.
4. Modify the shape type of drop buff item to 5.

Server/MirObjects/PlayerObject.cs
cjlaaa: The shape of the drop buff item I used before is 6, now it has been changed to 5.

Server/MirEnvir/Map.cs
```diff
@@ -1350,7 +1350,7 @@ private void CompleteMagic(IList<object> data)
                             SpellObject ob = new SpellObject
                                 {
                                     Spell = Spell.PoisonCloud,
-                                    Value = value + bonusdmg,
+                                    Value = value,// + bonusdmg,
```
Suprcode, line 1353: How come Stat.PoisonAttack (bonusDmg) is no longer used as part of PoisonCloud?

Client/MirScenes/GameScene.cs
```diff
@@ -6228,6 +6228,10 @@ public MirControl AttackInfoLabel(UserItem item, bool Inspect = false, bool hide
                 {
                     text = string.Format("Exp + {0}% ", minValue + addValue);
                 }
+                else if (realItem.Type == ItemType.Potion && realItem.Shape == 6)
```
Suprcode, line 6231: This condition also needs updating to shape == 5

Suprcode (changes requested):

cjlaaa:
Some modifications
1.PoisonCloud's green poison damage uses Stat.PoisonAttack.
2.Modified the shape type of the buff item to 5 on the client.

Server/MirEnvir/Map.cs
cjlaaa: I am sure that the official green poison damage is (MaxSC+MinSC)/2, but I am not sure if Stat.PoisonAttack is used. According to experience, PoisonAttack should be a bonus to green poison damage. I think the direct damage of poison cloud should not Receives a bonus, but the green poison damage caused by the poison cloud should have a bonus.The code has been modified.

Server/MirObjects/SpellObject.cs
```diff
@@ -140,7 +140,7 @@ public void ProcessSpell(MapObject ob)
                         if (ob.Dead) return;
 
                         if (!ob.IsAttackTarget(Caster)) return;
-                        ob.Attacked(((PlayerObject)Caster), Value, DefenceType.MAC, false);
+                        ob.Attacked(((PlayerObject)Caster), (Caster.Stats[Stat.MinSC] + Caster.Stats[Stat.MaxSC]) / 2, DefenceType.MAC, false);
```
Suprcode, line 143: Sorry. Just one final change needed.

This should still use "Value" - as you originally set Value in the PoisonCloud method "int damage = (Stats[Stat.MinSC] + Stats[Stat.MaxSC]) / 2;"

bonusDmg will still be separate as its another parameter so will not be applied here.

If you could fix this I can appy the merge.

Suprcode (changes requested):

Server/MirObjects/SpellObject.cs
cjlaaa: If "Value" is used here, the direct damage of the poison cloud will be boosted by Stat.PoisonAttack, I think it should only have a bonus to the green poison damage.But use "Value" here is ok too.

I confirm that the official poison cloud damage is (MaxSC+MinSC)/2. If you use magic.GetDamage, the damage of the PoisonCloud skill will be too large, I tested it.

Suprcode: Ah i see. Yes that makes sense not using "Value" here then for direct damage.

However, using the "magic.GetDamage" formula allows for much greater flexibility of how someone may want to use the skill.

If both direct damage and poison damage are both using "(MaxSC+MinSC)/2" then there is no possibility to change how much the spell will do in the files.

I suggest we leave the direct damage using the main "magic.GetDamage" formula (without the PoisonAttack) and have the poison damage use the formula "(MaxSC+MinSC)/2 + PoisonAttack"

cjlaaa: Okay, I forgot that "magic.GetDamage" can be modified directly in the database. I think your suggestion is the best way.

cjlaaa:
The skill PoisonCloud's direct damage use the main "magic.GetDamage" formula without the PoisonAttack bouns damage ,the green poison damage use the formula "(MaxSC+MinSC)/2 + PoisonAttack".

PR #462, opened 2021-10-04
Delete Character additional check.

JevLOMCN:


JevLOMCN:
[https://www.lomcn.org/forum/threads/delete-character-additional-check.106675/#post-1169647](url)

image

PR #469, opened 2021-12-19
Fix Spell.Plague (Taoist lv42 spell)

schen0x:
# Fix Spell.Plague (Taoist lv42 spell)

- fixed the randomization logic (previously wrong variables are used, hence poison will never trigger)
- fixed the affect area to 3x3 (previously affect 1x1 only)
- fixed the `tempValue` of poisonType red/green (previously the value of red and green seems reversed)
- Spell.Plague poison effect ignoreDefence = false -> true


Server/MirEnvir/Map.cs
```diff
@@ -820,6 +820,35 @@ public void Process(DelayedAction action)
             }
         }
 
+         /**
+         * return the coordinates of effect coordinates within an n x n square
+         * then use GetCell() in Map.cs to retrive real objects
+         * default 3x3
+         */
+        public static List<(int X, int Y)> GetEffectiveSquare(Point location, int mapWidth, int mapHeight, int squareEdgeLength = 3)
+        {
```
Suprcode, line 829: Please change this to use a List<Point> instead to keep it inline with existing code

Alternatively you could use the method "FindAllTargets" which will do the same thing but directly return a full list of target objects. Method would need to be copied from MonsterObject to PlayerObject.

Suprcode (changes requested):

Server/MirEnvir/Map.cs
schen0x: Thanks for the comment. Fixed as required.

PR #485, opened 2022-03-14
Lib Editor fix

JevLOMCN:


Lilcooldoode:
Can you speak to your friend again, these changes have been replaced for better versions (i.e. shortcut keys rather than clogging up the top menu with extra buttons)

PR #503, opened 2022-04-05
Change render logic

damianday:
Separate Drawing from Updating logic.


damianday:
This changes the rendering logic a lot, i've seen huge fps increases but it could do with more widespread testing.

PR #515, opened 2022-04-10
Fix #512

### LINKED ISSUE #512: New graphical glitch
Came across a random graphical glitch that causes a ghosting effect that doesn't disappear until you log out.

To make this occur without fail during testing, in Bichon Provice, teleport to a dark location on same map (e.g. 260 580) with no candle etc equipped, then teleport back to 330 260

I have a video showing this but it's too big to upload so a screenshot is the best I can do

image

damianday:


Lilcooldoode (commented):
I think there should be extra logic with this, have a look at RemoveObject packet, if the CellObjects list becomes empty it nulls it. 
Also need to think about TargetObject, MouseObject, MagicObject etc

Edit: Rather than simply clearing the TargetObject, MouseObject etc they should be updated so that if i am attacking a monster and then teleport the other side of it, i would continue to attack said monster

Knotty1985 (commented):
Updated: Logic works partially, it retains the attack on the spot you were hitting (where the monster was) even if its moved and a ghost is visible in the location where it was (around the player) until you attack another monster, initial issue does look to be resolved though, no NPC ghosting etc

damianday:
> RemoveObject

Going to have a look at the targetting, I've updated the pull for the Cells

PR #517, opened 2022-04-10
Update AMain.cs

damianday:
Made a start on making it more async friendly in order to improve upload.
This is experimental until tested! but the changes are sane.
What remains to be done is File.ReadAllBytes can block.
Motivation for this pull is the UI locking up and being unresponsive while uploading.

### DIFF
--- a/AutoPatcherAdmin/AMain.cs
+++ b/AutoPatcherAdmin/AMain.cs
@@ -280,7 +280,6 @@ namespace AutoPatcherAdmin
             session.Open(sessionOptions);
         }
 
-
         private void BeginUpload()
         {
             if (UploadList == null) return;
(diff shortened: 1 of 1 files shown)


AutoPatcherAdmin/AMain.cs
```diff
@@ -725,12 +734,18 @@ private byte[] Compress(byte[] raw)
...
+            using (GZipStream gStream = new GZipStream(stream, CompressionMode.Compress, true))
+                gStream.WriteAsync(raw, 0, raw.Length);
+        }
     }
 
     public class FileInformation
     {
         public string FileName; //Relative.
-        public int Length, Compressed;
+        public long Length, Compressed;
```
damianday, line 748: This would be a breaking change I assume..

PMySQLQ:
When uploading, the client files go to disk D

Suprcode:
> When uploading, the client files go to disk D

The client files for temp upload should be copied to the same directory as you're running the admin patcher from.

Suprcode:
Closing this as no response to updating. Feel free to reopen once the suggested change has been made.

PR #519, opened 2022-04-11
Undo last target

damianday:
There are a few bugs with this change.


Lilcooldoode:
What are the bugs? Havent seen any reported?

damianday:
I had two players in border village, have players target each other, then teleport to bichon wall and teleport back to border village with each player, it should create a ghost.

PR #533, opened 2022-05-29
Mon357 Mon387

angelk727:


angelk727:
[](url
[387.zip](https://github.com/Suprcode/mir2/files/8792488/387.zip)
)

angelk727:
[](url
[357.zip](https://github.com/Suprcode/mir2/files/8792498/357.zip)
)

zhaokai1982:
Directly log in to your git account on your visual studio, and then pull this example. After modifying and writing the code, you can synchronize the code,

Suprcode:
The crystal beast frames are all incorrect, none of them match up to the images.
The cheiftan frame are all incorrect. none of them match up to the images.

You've also added blank spaces in the libs. Please remove all blank spaces (using the remove blank spaces option) to reduce the file size.

image

Cannot merge code until these lib's are fixed.

angelk727:
Ok ok, please ignore my submission. I just want to say that if you don't use blank pictures in the monster stealth phase, you can't make an effect! There are also 10 action pictures of monster attack and only 5 special effect pictures. For example, 0 represents an empty 1 represents a special effect picture. 0011111000 can I express this without using blank pictures 1111100000 or 0000011111? What you said is also true. Blank pictures occupy some resources, but do not occupy the memory space of the server

zhaokai1982:
I deleted the blank image. Only in this way can I attack this monster in three forms. There is no remote attack. Delete the blank image and adjust the code reading rules. Thank you very much.

zhaokai1982:
The blank image of the attack effect should also be removed.

Suprcode:
> Ok ok, please ignore my submission. I just want to say that if you don't use blank pictures in the monster stealth phase, you can't make an effect! There are also 10 action pictures of monster attack and only 5 special effect pictures. For example, 0 represents an empty 1 represents a special effect picture. 0011111000 can I express this without using blank pictures 1111100000 or 0000011111? What you said is also true. Blank pictures occupy some resources, but do not occupy the memory space of the server

If the monster has a "stealth phase" (i'm guessing you mean invisible) then this should be done in the server code by removing the object from the map (just like how player transparency or observer works). This shouldn't be done by faking blank frames.

If the action has more frames than the magic, then you can either use an Effect object which starts after a certain frame / duration.

image

Or if the effect matches up with the action in all directions, you can use Draw, and just draw it at the same time as the action frame.

image


There are 500+ examples of other monster's and AI which you can look at for examples. There should be no need to add blank frames.

PR #552, opened 2022-08-07
Fix Middle Layer Tile strips Draw for Wemade & Shanda Mir2 Libraries.…

mir2pion:
… M2P


Suprcode:
Are you sure this doesn't break drawing on any existing mir3 maps such as old EI or JoB versions? This check will have been added originally to purposely exclude certain indexes.

Also your pull request has checked in several additional files, these just need excluding before submitting it please so only the code changes are brought through.

mir2pion:
This is what happened - I made a fork of mir2 and downloaded new mir2 master and it opened in VS with code having red underline squiggles all over. I searched and found that deleting .vs folder from the solution will fix things and the folder will regenerate next time VS starts... 

I was more careful, instead of deleting it I appended .BAK 'extension' to it (.vs.BAK) and left it among solution files. 
VS generated new .vs file and the red underlines thing was gone but VS still kept track of this .BAK file which I didn't expect. 

It even got pushed to my GH repo which I didn't fail to notice but doing something about is beyond me, I figured if I attempt to get rid of it, I will probably mess things up even more.

How about deleting this pull request and I redo the whole thing from the very start again. It should be easy enough and a bit of practice will only be good for me

PR #556, opened 2022-08-14
UI fixes

OuYangPlane:
Fixed the MIR2UI framework closer to WEMADE, and canceled the bracketed attribute of attack speed, more like the attribute layout of the official server

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -7341,7 +7341,7 @@ namespace Client.MirScenes
                 {
                     string negative = "+";
                     if (addValue < 0) negative = "";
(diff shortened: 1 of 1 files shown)


Suprcode:
You've completely removed the added stats display from item labels. If that's what official mir2 korea is like I think we'd prefer to not have it match. Certain things don't need to match 100%, especially when we've done it a better way.

Feel free to post this online lomcn to ask other peoples opinions, but I think people would generally prefer to see how much a stat has increased from default.

The border colour changing when dura = 0 is a good idea though, feel free to create a new pull request with just this addition.

lodos3:
🤣

PR #570, opened 2022-08-29
Better font size for resolution

OuYangPlane:
Better font size for resolution


Suprcode:
The fontsize is controlled by the client settings "FontSize".

You should update this to use Settings.FontSize instead - so people can choose to make the font size whatever they want.

PR #576, opened 2022-09-06
Map Range Fix

KingdomMir2:
added map range draw fix.


Suprcode:
Please remove the colour change for guild names - i don't think that should be there.

Your branch looks to be out of date as you're trying to merge multiple previous checkins too.

PR #586, opened 2022-10-17
[BUG FIX] Crafting Gold Count 

EliteMir:
Crafting multiples was not taking the gold cost correctly and had a missing count on removal.


EliteMir:
Making adjustments

PR #587, opened 2022-10-17
[BUG FIX] Crafting NPC to take the correct amount of gold based on quantity of items crafted

EliteMir:


EliteMir:
Only need the last commit pushing, rest can be removed as they aren't doing anything.. not sure how to get rid of them from the pull request.

PR #590, opened 2022-11-09
[CRASH FIXES] IsAttackTarget & ProcessSpell checks

EliteMir:
IsAttackTarget is allowing null objects to pass through to spells.


EliteMir:
Added another crash fix for Blizzard

PR #595, opened 2022-11-29
DoS / Packet Flood Protection

EliteMir:
- Limits client connections
- Prevents packet flooding to the server client port
- DoS protection

Currently running live on our server and tested.

Enjoy!
**Team Elite**


Suprcode:
Added manually.

PR #625, opened 2023-01-29
Update rebirth effects

OuYangPlane:
The feature file can directly replace the client file. Except for the addition of rebirth effects, nothing has changed. Clothes in the server can call these effects effectlink : https://drive.google.com/file/d/104K0wroDyiwAgUjsxGq-AZQn7zieyAee/view?usp=sharing Please upload it to the server so that the client can update


Suprcode:
You should add these effects under the "LevelEffects" enum instead.

Then you can add them under a specific flag number - which people can use however they want (such as Rebirth).

It shouldn't be added as item effects.

PR #671, opened 2023-02-28
Mysterious Moon Forest Renewal Humanoid Monster

PMySQLQ:
Mysterious Moon Forest Renewal Humanoid Monster

### DIFF
--- a/Client/Client.csproj
+++ b/Client/Client.csproj
@@ -44,7 +44,7 @@
     <DebugSymbols>true</DebugSymbols>
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
-    <OutputPath>..\Build\Client\</OutputPath>
+    <OutputPath>E:\Client\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>1</WarningLevel>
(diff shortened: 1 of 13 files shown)


Suprcode (changes requested):

Suprcode:
If you could update the wiki on lomcn with the new AI numbers for these monsters that would be appreciated.

Suprcode:
Are there any new monster libs which are needed? or are all monsters based off human appearance?

MousaEqilan:
> based off human appearance

i think he add as based off human appearance as there no line for them in Shared/Enums.cs

PMySQLQ:
No monster library required

Fill in the AI number

               case 214:
                    return new SepWarrior(info);
                case 215:
                    return new SepWizard(info);
                case 216:
                    return new SepTaoist(info);
                case 217:
                    return new SepAssassin(info);
                case 218:
                    return new SepArcher(info);

                case 219:
                    return new SepHighWarrior(info);
                case 220:
                    return new SepHighWizard(info);
                case 221:
                    return new SepHighTaoist(info);
                case 222:
                    return new SepHighAssassin(info);
                case 223:
                    return new SepHighArcher(info);

EliteMir:
Do you mean the forest in Tao Village?

Suprcode:
> Do you mean the forest in Tao Village?
> 
> Also you need to list the lib numbers of what these mobs are supposed to be.

Looking at the code I think you can just leave them as a Hen - the AI takes over the images and applies player armour/weapons to them.

Suprcode:
Why close? @PMySQLQ

EliteMir:
Really nice AIs... although I'm not sure if these are actually real?

https://www.mir2db.com/list/127

These don't exist on mir2db

PR #701, opened 2023-03-02
Add new Assassin skill CatTongue

PMySQLQ:
Add new Assassin skill CatTongue

Has 2 negative states

image


PMySQLQ:
Add new Assassin skill MoonMist
Stealth animation damage after using skills


image

PR #718, opened 2023-03-20
Fix Item Move Error

MousaEqilan:
when move items sometime with hero bag or normal bag will do error
this should fix it as i test and work fine for me


Suprcode:
Why would increasing the player slots by 2 fix a move error?

Players have 40 slot inventory + 6 belt slots. So it should be 46 slots.
Heroes have 40 slot inventory + 2 belt slots. So it should be 42 slots.

I'm not sure what the error is as you haven't linked anything - but what ever it is, i don't think expanding the inventory size is correct.

MousaEqilan:
the error was here 
https://www.lomcn.org/forum/threads/public-test-server.109970/post-1191427
also this been fixed at my code from few weeks and till now not have this move item error anymore 
the move items from both side sometime make this error 
also aren't both idx is 8 so should be 40+8= 48 not 46 that been bug from original code 
when you not use heroes at your server then size should be 0 if you keep set to 46

Suprcode:
why would idx be 8? You don't have 8 belt slots. The player and hero slots are separate so you don't just add them together.

just because you dont have the error anymore it doesn't mean you've fixed it correctly.

Sounds to be like theres a bug with the slot number being different being sent from the client, so it needs to take away of hero belt idx or something.

MousaEqilan:
aha i got you now will do more test and see where make me going too

PR #726, opened 2023-03-30
Various Issues Resolved

grimchamp:
**[Issue-644](https://github.com/Suprcode/mir2/issues/644)**
- Curse now causes attacking player to go brown.
- Poison action was not being assigned an owner for browntime calculation.

**[Issue-645](https://github.com/Suprcode/mir2/issues/645)**
Attacking a players pet with TurnUndead will now turn you brown if pet owner:
- is not red
- is not brown
- is not at war

**[Issue-648](https://github.com/Suprcode/mir2/issues/648)**
Enable HP/MP regen when running.

**[Issue-680](https://github.com/Suprcode/mir2/issues/680)**
- Lover buff once set would never take new server values if changed.
- Now removes on log out to reapply.

**[Issue-686](https://github.com/Suprcode/mir2/issues/686)**
- Shoulder dash could push target from any distance not just directly in front of player. This is now working properly.
- Refactored logic to be more efficient and readability for debugging.

**[Issue-687](https://github.com/Suprcode/mir2/issues/687)**
- Focus Master Target pet mode added.
- Keybind added to allow for shortcut. By default has none set.
- Server now stores master target (previously only stored client side).

**[Issue-688](https://github.com/Suprcode/mir2/issues/688)**
Teleporting on same map retains current target.

**[Issue-690](https://github.com/Suprcode/mir2/issues/690)**
- Fix flaming sword not having cooldown timer on hotbar.
- CastTime was never being set.

**[Issue-693](https://github.com/Suprcode/mir2/issues/693)**
- Fixed light offset for Blizzard and Meteor (previously map was being lit many tiles above spell location)
- Underlying issue: animation Y co-ord was being changed due to size of animation and thus pushed out light by Y-20.
- New AnimationOffset Point in SpellObject allows for offset to animation frames.

**[Issue-695](https://github.com/Suprcode/mir2/issues/695)**
- Added setting in keybinds under 'Combat'. This setting is not set by default and when held down will use target location rather than cursor for following spells:

Firebang
MassHiding
SoulShield
Firewall
MassHealing
Blizzard
MeteorStrike
TrapHexagon
Plague
ArcherSummonStone
HealingCircle

- Changing a setting in keybind screen,  now saves to Keybinds.ini - previously did not save to file.


MousaEqilan:
https://github.com/Suprcode/mir2/issues/645
this code not at main source at all i think there your custom, code need edit also

PR #741, opened 2023-04-03
Issue 660 - Various New Item Form Improvements

grimchamp:
[Issue 660](https://github.com/Suprcode/mir2/issues/660)

- Fixed issue where changing a dropdown value did not mark row as modified. Example field 'Type'.
- Fixed issue where changing a checkbox value did not mark row as modified. Example field 'Start Item'.
- Fixed an issue in cellvalidating that was stopping text input cells from being marked as modified.
- Fixed a condition where a silent NullReferenceException was being thrown in UpdateFilter() on form load.
- "Modified" column is now readonly to avoid accidental cancellation of saves.
- Trying to close form with errors will make popup appear listing errored items (index number and item name). Pressing 'OK' will discard those specific changes and save rest of modified fields whilst pressing 'Cancel' will take you back to form.


Server.MirForms/Database/ItemInfoFormNew.cs
```diff
@@ -918,5 +927,55 @@ private void Gameshop_button_Click(object sender, EventArgs e)
...
+                }
+            }
+
+            if (inError.Count > 0)
+            {
+                String msg = string.Join(Environment.NewLine, inError);
+                if (MessageBox.Show($"The following items are invalid: {msg}", "Discard Invalid Items?", MessageBoxButtons.OKCancel) == DialogResult.OK)
+                {
+                    SaveForm();
+                    Envir.SaveDB();
```
Suprcode, line 968: i think you want a return here? Otherwise it'll save twice

Suprcode (changes requested):

Server.MirForms/Database/ItemInfoFormNew.cs
grimchamp: Change committed.

PR #743, opened 2023-04-05
Rebirth effect add (1)

OuYangPlane:
file https://drive.google.com/file/d/104K0wroDyiwAgUjsxGq-AZQn7zieyAee/view?usp=sharing


JevLOMCN:
The effect is being utilised by detecting a Flag.

what are the flag numbers for each rebirth?

Suprcode:
Happy to accept this - but your PR (2) needs a small modification before I can.

PR #744, opened 2023-04-05
Rebirth effect add (2)

OuYangPlane:


Shared/Enums.cs
```diff
@@ -87,10 +87,16 @@ public enum AwakeType : byte
...
+     None = 0,
+    Mist = 1,
+    RedDragon = 2,
+    BlueDragon = 4,
+    Rebirth1 =8,
+    Rebirth2 = 16,
+    Rebirth3 = 32,
+    NewBlue = 64,
+    YellowDragon = 128,
+    Phoenix= 144,
```
Suprcode, line 99: 144 is wrong. The next number should be 256 - which means you'll need to change this LevelEffects to a ushort, and also change the packets to accept ushort too.

Suprcode (changes requested):

PR #784, opened 2023-04-15
Better performance of special effects

OuYangPlane:
Fixed the timing of the sole effect,
fixed the offset of the rebirth 2 effect（file needs to be replaced https://drive.google.com/file/d/104K0wroDyiwAgUjsxGq-AZQn7zieyAee/view?usp=share_link）

### DIFF
--- a/Client/MirObjects/PlayerObject.cs
+++ b/Client/MirObjects/PlayerObject.cs
@@ -665,11 +665,10 @@ namespace Client.MirObjects
             if (LevelEffects.HasFlag(LevelEffects.BlueDragon))
             {
                 Effects.Add(new SpecialEffect(Libraries.Effect, 1210, 20, 3200, this, true, true, 1) { Repeat = true });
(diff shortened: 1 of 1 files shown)


Suprcode:
The libs should be directly from official, so why are the offsets incorrect in the first place?

OuYangPlane:
> The libs should be directly from official, so why are the offsets incorrect in the first place?

This is a problem that cannot be ignored, he does have such an offset

OuYangPlane:
> The libs should be directly from official, so why are the offsets incorrect in the first place?

If you have doubts about the offset of Rebirth 2, you can pass this code first, because the code only changes the playback effect of the animation (he is better), not the coordinate offset

Suprcode:
I've just checked the official Effect.WIL file, and none of those effects exist there.

Where are these effects located in the official client?

OuYangPlane:
> I've just checked the official Effect.WIL file, and none of those effects exist there.
> 
> Where are these effects located in the official client?

D:\Wemade\Mir2Global\Data\Magic3.wil

Suprcode:
Please leave this open. The code will be redone correctly to look at the Magic3

PR #833, opened 2023-06-10
Rebirth effect adjustment

OuYangPlane:
Modify the rebirth effect to read the MAGIC3.lib file
Readjust the animation playback state
Some files in the client need to be replaced to the public download server (they are from the latest Korean official, without any modification, and the difference with the crystal has been checked, the item order is the same as the crystal, and I will update the database with JEV later)
https://drive.google.com/file/d/1CNX6kyWlIHbtYLb-Ci3jtclLQD5ygWb5/view?usp=sharing

### DIFF
--- a/Client/MirObjects/PlayerObject.cs
+++ b/Client/MirObjects/PlayerObject.cs
@@ -685,47 +685,47 @@ namespace Client.MirObjects
 
             if (LevelEffects.HasFlag(LevelEffects.Rebirth1))
             {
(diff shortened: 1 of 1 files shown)


Suprcode:
i'll accept this PR and will manually download the magic files from official client.

If there are other files which need updating with missing official images please raise a separate issue ticket. Crystal does not use a single npc file, and has custom buff icons - so these cannot just be added to the client.

PR #837, opened 2023-06-22
Maintain 800 resolution

OuYangPlane:
1. Fixed the dislocation of the hero's 800 resolution and used the IF statement to solve the conflict between 1024 resolution and 800 resolution
2,800 resolution is a nostalgic resolution, and its effect has an irreplaceable position among the old MIR players for many years


Suprcode:
No sorry, 800 theme was purposely removed due to needing too much custom support

OuYangPlane:
I will be responsible for maintaining this resolution, please pass it

PR #839, opened 2023-06-30
MonGen With Range

JevLOMCN:
#ACT
MONGEN MOB COUNT INSTANCE RANGE

EXAMPLE:

#ACT
PARAM1 n0
PARAM2 300
PARAM3 300
MONGEN Deer 5 0 100


Suprcode:
This PR has multiple coding issues. Please correct then submit a new PR

PR #840, opened 2023-06-30
NPC/Quest CSV format.

JevLOMCN:
Both NPC and Quest Import/Export changed to use CSV file format.

### DIFF
--- a/Server.MirForms/Database/QuestInfoForm.cs
+++ b/Server.MirForms/Database/QuestInfoForm.cs
@@ -6,7 +6,7 @@ namespace Server
 {
     public partial class QuestInfoForm : Form
     {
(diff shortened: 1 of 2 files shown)


Server.MirForms/Database/QuestInfoForm.cs
```diff
@@ -226,7 +226,7 @@ private void ImportButton_Click(object sender, EventArgs e)
             string Path = string.Empty;
 
             OpenFileDialog ofd = new OpenFileDialog();
-            ofd.Filter = "Text File|*.txt";
+            ofd.Filter = "CSV File" +"*.csv";
```
Suprcode, line 229: Small error here - should be "CSV File|*.csv"

Suprcode (changes requested):

JevLOMCN (commented):
Sorted

PR #841, opened 2023-07-01
Fixes for cross-platform support

n7st:
# Overview

This pull request adds support for building the Server library on Linux. It seems that for whatever reason `net7.0-win` automatically imports `System.Drawing`, which is required for the `Point` struct used in a lot of the changed files below. The Server library also doesn't contain any WinForms code, so those can be disabled.

I've successfully built the server on both Windows and Linux.

## Windows build output

This is the full Server.MirForms application:

image

And just to demonstrate the Windows build is still working correctly (ignore the map errors, I'm working with a random assortment I found on an old PC):

image

## Linux build output

This is just the Server library:

image

# Changelog

## Changed

* The Server library now targets `net7.0` instead of `net7.0-win`.
* The Server library no longer requires WinForms to build.

## Fixed

* Restored `System.Drawing` imports in files which use the `Point` struct.
* Removed an unused variable.

### DIFF
--- a/Server/MirObjects/Monsters/SepHighWizard.cs
+++ b/Server/MirObjects/Monsters/SepHighWizard.cs
@@ -30,8 +30,6 @@ namespace Server.MirObjects.Monsters
                 return;
             }
 
-            DelayedAction action;
-
             ShockTime = 0;
             ActionTime = Envir.Time + 300;
             AttackTime = Envir.Time + AttackSpeed;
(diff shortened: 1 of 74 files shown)


Server/MirObjects/Monsters/SepHighWizard.cs
```diff
@@ -30,8 +30,6 @@ protected override void Attack()
                 return;
             }
 
-            DelayedAction action;
```
n7st, line 33: This variable was unused.

PR #843, opened 2023-07-10
Fix for NPC scripts

JevLOMCN:
Accidently changed the NPC script's to expect CSV instead of TXT.


MousaEqilan:
When you click NPC from the database, a new, empty CSV file will open for you
when you delete the original TXT file, NPC will not work with you in the game

PR #881, opened 2023-12-14
Minimap Preview

JevLOMCN:
MapInfoForm now displays a Minimap preview.
Minimaps folder will auto create if not present, you can find this folder inside the Server's Envir folder.

Tested as working
[Image](https://imgur.com/tenh3i2)


Suprcode:
I don't see the code to create the folder if it doesn't exist? I believe you'll need to add something in the settings.cs to do that

Regardless, will accept as is and you can open a new PR if that bit of code is missing

JevLOMCN:
Server.Library> Settings.cs

31: MinimapsPath = Path.Combine(EnvirPath, "Minimaps");

The code is there.

PR #885, opened 2023-12-18
NPC Preview

JevLOMCN:
NPCInfoForm now displays a NPC Image preview.
NPCPreview folder will auto create if not present, you can find this folder inside the Server's Envir folder.

### DIFF
--- a/Server/Settings.cs
+++ b/Server/Settings.cs
@@ -28,7 +28,8 @@ namespace Server
             NameListPath = Path.Combine(EnvirPath, "NameLists"),
             ValuePath = Path.Combine(EnvirPath, "Values"),
             NoticePath = Path.Combine(EnvirPath, "Notice.txt"),
-            MinimapsPath = Path.Combine(EnvirPath, "Minimaps");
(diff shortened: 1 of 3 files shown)


Server/Settings.cs
```diff
@@ -28,7 +28,8 @@ private static MessageQueue MessageQueue
             NameListPath = Path.Combine(EnvirPath, "NameLists"),
             ValuePath = Path.Combine(EnvirPath, "Values"),
             NoticePath = Path.Combine(EnvirPath, "Notice.txt"),
-            MinimapsPath = Path.Combine(EnvirPath, "Minimaps");
+            MinimapsPath = Path.Combine(EnvirPath, "Minimaps"),
+            NPCPreviewPath = Path.Combine(EnvirPath, "NPCPreviewPath");
```
Suprcode, line 32: Can you move this and the minimap path in to a sub folder of "Previews". And also rename "NPCPreviewPath" to just "NPCs"

Suprcode (changes requested):

Server/Settings.cs
JevLOMCN: Yea can do.

PR #893, opened 2024-01-01
Flag/Quest ID Checker

JevLOMCN:
New Search Box added to PlayerInfoForm, This will allow server owners to search if a player has a Flag set as active or inactive.

Also added Map Title (Name) to display with Current Location

### DIFF
--- a/Server.MirForms/Account/PlayerInfoForm.cs
+++ b/Server.MirForms/Account/PlayerInfoForm.cs
@@ -39,7 +39,7 @@ namespace Server
 
             if (Character.Player != null)
                 CurrentMapLabel.Text =
-                    $"{Character.Player.CurrentMap.Info.FileName} {Character.CurrentLocation.X}:{Character.CurrentLocation.Y}";
+                    $"{Character.Player.CurrentMap.Info.Title} {Character.Player.CurrentMap.Info.FileName} {Character.CurrentLocation.X}:{Character.CurrentLocation.Y}";
             else
                 CurrentMapLabel.Text = "OFFLINE";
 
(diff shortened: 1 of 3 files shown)


Server.MirForms/Account/PlayerInfoForm.cs
```diff
@@ -133,5 +133,122 @@ private void OpenAccountButton_Click(object sender, EventArgs e)
...
+                if (questID < 1)
+                {
+                    QuestResultLabel.Text = "Invalid Quest ID";
+                    QuestResultLabel.ForeColor = Color.Red;
+                    return;
+                }
+
+                bool isQuestComplete = Character.CompletedQuests.Contains(questID);
+
+                QuestResultLabel.Text = $"Quest {questID} is {(isQuestComplete ? "Complete" : "Not complete")}";
```
Suprcode, line 218: maybe needs to be abit more sophisticated for quest lookup. As well as completed also check active quests.

Server.MirForms/Account/PlayerInfoForm.cs
```diff
@@ -133,5 +133,122 @@ private void OpenAccountButton_Click(object sender, EventArgs e)
...
+
+            if (int.TryParse(flagTextBox.Text, out int flagIndex))
+            {
+                if (flagIndex >= 0 && flagIndex < Character.Flags.Length)
+                {
+                    bool flagValue = Character.Flags[flagIndex];
+
+                    if (flagValue)
+                    {
+                        ResultLabel.Text = $"Flag {flagIndex} ACTIVE.";
```
Suprcode, line 154: change "ACTIVE." to "is active"
change "INACTIVE." to "is inactive"

so it matches the output format of quests

Suprcode (changes requested):

Server.MirForms/Account/PlayerInfoForm.cs
JevLOMCN: This has been adjusted now

Server.MirForms/Account/PlayerInfoForm.cs
JevLOMCN: Not sure how to change the code, I know that 0 is not complete and 1 is complete.

JevLOMCN:
All adjustments have been made.

PR #909, opened 2024-01-24
MassHealing Heroes Fix

JevLOMCN:
Code now allows MassHealing to work on Heroes


JevLOMCN:
> May be MassHiding also need to be changed.

I'll adjust it and pr again.

PR #914, opened 2024-02-19
Update IntelligentCreatureObject.cs

damianday:
Proper null checking for TargetList.
PickupAllItems - use location.
Guard against null cells (not likely..)


Suprcode:
no need for any cell null checks - its impossible for them to be null at its a grid array. Anything out of bounds would throw an exception before your null check - and nothing in bounds is ever null

damianday:
I've removed the cell null checks.

PR #946, opened 2024-05-12
Fix large NPC goods

damianday:
Can cause considerable lag when retrieving items from npc, break the lists down into manageable chunks.


damianday:
Another option would be to just send each item as NewItem with a flag to indicate it's for the NPC not user

PR #947, opened 2024-05-13
Fix #667

### LINKED ISSUE #667: Goods System - Incorrectly coded
Was speaking to Far about this and the code seems incorrect.

'Max Goods Stored' if this is set to 10.. it is not 10 items..

It is 10 items... of THAT item.. of THAT item type.
Meaning if you have 100 weapons..
10 x 100 per person

It should be 10 TOTAL items combined for all.

This will cause your server to lag massively if the goods file excesses a certain size as it reloads in the process every hour.

damianday:


Suprcode:
I dont think this is an ideal fix for the issue as it basically breaks the used goods system. If you change the system to just overall max quantity, then you have all the later items in the list have no used goods at all.

The original idea was allowing a max quantity of each item type (IronSword, WoodenSword etc) to have a maximum quantity of used versions which can be stored. By default i think its set as max 10 per item.

The issue is if you have a server which has 100 different weapons, and they're all being sold to this 1 npc, the npc now needs to send 100x10 items each request.

Your previous fix might have aleviated the issue with chunks - if not i think the ideal fix would be to add paging to the used goods and only send ones in view.

damianday:
I think the paging idea is good, but maybe complicated to add due to the way the client organises and displays the goods.
8 items per page, but each item could have many sub items.
I think an easier solution would be to compress the items and just send the compressed bytes in chunks, what do you think?

Suprcode:
Your previous change has chunked the goods sending for new and used goods - so hopefully that in itself will be enough to fix the lag without the need to compress it.

I'd say leave it as it is for now and see if large amounts of used goods are still causing issues.

damianday:
Ok 👌 
I'll close this pull request, since the fix is invalid.

PR #949, opened 2024-05-16
Weather System

JevLOMCN:
Massive thanks to Win & TheDeath for helping.

There will be some further commits.

Database Version 110 Please back up your database before updating.


Jay-lee-01:
When leaving or entering a store, the client quits. I think it's a bug.
(For example, the client terminates when entering a clothing store from Bicheonseong or leaving a clothing store)

PR #958, opened 2024-05-31
Some Fixes

meacher0:


Server/MirObjects/PlayerObject.cs
```diff
@@ -12978,15 +12978,14 @@ public void GameshopBuy(int GIndex, byte Quantity)
...
                 
                 var cost = Product.CreditPrice * Quantity;
-                if (cost < Account.Credit || cost == 0)
+                if ((cost < Account.Credit || cost == 0) && PType == 0)
                 {
                     canAfford = true;
                     CreditCost = cost;
                 }
-                else if (Product.CanBuyGold)
+                else if (Product.CanBuyGold || PType == 1)
```
Suprcode, line 12986: This will allow people to buy any item with gold - even if that item is only marked as credit only.

You also look to have removed the ability to buy items with a mixture of credit and gold.

Unsure what issue you're trying to fix here?

Suprcode (changes requested):

Server/MirObjects/PlayerObject.cs
meacher0: The Client only specifies "Buy with Gold" or "Buy with Credits" which is confusing players. At the moment, if you buy with credits it will just take credits. if you buy with gold and have any credits, it will first take the credits and then gold. 

so if you select buy with gold, it costs you the credits anyway, and if you don't have enough credits, it will take them all. so you're just choosing to pay extra gold for no reason?

Suprcode: im fine with removing the ability to pay with both when you don't have enough of the one you want - and instead just say "insufficient amount", if that's less confusing. I think the client side also needs updating so it doesn't offer this too though?

But your check needs changing to an & so it can't be exploited by the client.

meacher0: So, i need to update that line above to "else if (Product.CanBuyGold && PType == 1)" instead of using || ?
Do i just do an extra commit for that change?
(Sorry, still very much learning)

meacher0: Ahhh s**t yeah, i think i see it, needs to be "&&" because otherwise it can bypass the CanBuyGold check...
i think see what ya did there...

Suprcode: That whole if block (between 12980 and 13000) ideally wants fully rewriting as its overly complicated.

just wants to be 

if (PType == 0)
//check can buy with credit and can afford
else if (PType == 1)
//check can buy with gold and can afford
else
//invalid type passed, return and log

meacher0:
image
Like this?

Server/MirObjects/PlayerObject.cs
meacher0: image
Like this?

meacher0: miseed the CanBuyGold and Credit check thing...
soooo...
This?
image

meacher0: i can remove the cost == 0 and goldcost == 0 parts cant I ? 
because if the price is 0 then its less than whats available so all good but still needs checking if CanBuy...

PR #967, opened 2024-06-06
Heal Hero in SafeZone

meacher0:


Jay-lee-01:
There is one problem with this. In the safe zone, the hero gets healed, but the player doesn't get healed.

PR #973, opened 2024-06-18
Removed Tree AI's "attacked" logic from RedMoonEvil

meacher0:


Suprcode:
Why?

meacher0:
It appears that the Tree's AI was used as a base for RME and the "Attacked" code was left in. After checking with @OuYangPlane and asking in the LOMCN:Mir2 Discord, it was confirmed that not only should RME take damage normally, it should also be susceptible to negative buffs, both of which the Tree AI's "attacked" logic prevents.

PR #979, opened 2024-07-05
Character List

JevLOMCN:
Added a Character List which will display all Characters.

Before this the process of searching Characters was to either go through them individually using the Account Info Form or the Players Online Tab.

Also has a Filter By Item UID box which allows you to quickly identify the owner of an Item.

### DIFF
--- a/Server.MirForms/SMain.cs
+++ b/Server.MirForms/SMain.cs
@@ -565,8 +565,6 @@ namespace Server
             CharacterInfoForm form = new CharacterInfoForm();
 
             form.ShowDialog();
-
-            form.Dispose();
         }
     }
 }
--- a/Server.MirForms/SMain.cs
+++ b/Server.MirForms/SMain.cs
@@ -1,4 +1,5 @@
 ﻿using CustomFormControl;
+using Server.Account;
 using Server.Database;
 using Server.MirDatabase;
 using Server.MirEnvir;
@@ -557,6 +558,15 @@ namespace Server
             SystemInfoForm form = new SystemInfoForm(8);
 
             form.ShowDialog();
+        }
+
+        private void CharacterToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            CharacterInfoForm form = new CharacterInfoForm();
+
+            form.ShowDialog();
+
+            form.Dispose();
         }
     }
 }
--- a/Server/MirEnvir/Envir.cs
+++ b/Server/MirEnvir/Envir.cs
@@ -124,6 +124,7 @@ namespace Server.MirEnvir
         public int NextAccountID, NextCharacterID, NextGuildID, NextHeroID;
         public ulong NextUserItemID, NextAuctionID, NextMailID, NextRecipeID;
         public List<AccountInfo> AccountList = new List<AccountInfo>();
+        public List<AccountInfo> ArchivedAccountList = new List<AccountInfo>();
         public List<CharacterInfo> CharacterList = new List<CharacterInfo>();
         public List<GuildInfo> GuildList = new List<GuildInfo>();
         public LinkedList<AuctionInfo> Auctions = new LinkedList<AuctionInfo>();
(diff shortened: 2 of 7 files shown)


Server/MirEnvir/Envir.cs
```diff
@@ -2532,7 +2533,93 @@ public bool CharacterExists(string name)
...
+            }
+
+            return list;
+        }
+        public List<CharacterInfo> MatchPlayerbyItem(string ItemID, bool match = false)
+        {
+            if (string.IsNullOrEmpty(ItemID)) return new List<CharacterInfo>(CharacterList);
+
+            ulong itemId = 0;
+            ulong.TryParse(ItemID, out itemId);
```
Suprcode, line 2563: you should probably just return nothing if the id isn't a number

Server/MirEnvir/Envir.cs
```diff
@@ -2532,7 +2533,93 @@ public bool CharacterExists(string name)
...
+
+            List<CharacterInfo> list = new List<CharacterInfo>();
+
+            for (int i = 0; i < CharacterList.Count; i++)
```
Suprcode, line 2567: i dont get the difference between match and not a match?

Server.MirForms/SMain.cs
```diff
@@ -558,5 +559,14 @@ private void heroesToolStripMenuItem_Click(object sender, EventArgs e)
...
             form.ShowDialog();
         }
+
+        private void CharacterToolStripMenuItem_Click(object sender, EventArgs e)
+        {
+            CharacterInfoForm form = new CharacterInfoForm();
+
+            form.ShowDialog();
+
+            form.Dispose();
```
Suprcode, line 569: why are you disposing it straight after showing it?

Server/MirEnvir/Envir.cs
```diff
@@ -2532,7 +2533,93 @@ public bool CharacterExists(string name)
...
+                    foreach (var item in CharacterList[i].AccountInfo.Storage)
+                        if (item != null && item.UniqueID == itemId && !list.Contains(CharacterList[i]))
+                            list.Add(CharacterList[i]);
+                }
+
+            }
+
+            return list;
+        }
+        public AccountInfo GetAccountArchived(string accountID)
```
Suprcode, line 2615: method isn't called by anything?

Suprcode (changes requested):
Several incomplete features and weird code issues to fix here. No need to rush pull requests through until it 100% works as intended.

Also please provide screenshots when forms are changed/added so its easier to review

Server.MirForms/SMain.cs
JevLOMCN: Fixed

Server/MirEnvir/Envir.cs
JevLOMCN: Removed

Server/MirEnvir/Envir.cs
JevLOMCN: I've just removed the null check. If a letter is input then it returns nothing and if a unused item id is input it does the same.

Server/MirEnvir/Envir.cs
JevLOMCN: When match is set to false, the method does not check:

AccountInfo.Storage: Excludes checking items stored in the account storage.
Mail: Excludes checking items attached to mails.

Server/MirEnvir/Envir.cs
Suprcode: yes but why bother doing the rest of the code when you know 0 will never find anything?

if its not parsed as a number, or if 0 is entered - it should just return no results straight away

JevLOMCN: I see, I'll do a pr

Server/MirEnvir/Envir.cs
Suprcode: In that case just wrap the check around those loops which the "match" utilises, so you don't have to duplicate the entire code

Server/MirEnvir/Envir.cs
Suprcode: The archived list at the top also needs removing please as that seems to have no use

Server/MirEnvir/Envir.cs
JevLOMCN: image

Server/MirEnvir/Envir.cs
JevLOMCN: Fixed

Server/MirEnvir/Envir.cs
JevLOMCN: I have no clue how

Server/MirEnvir/Envir.cs
Suprcode: Move the list above it and just return the empty list

Server/MirEnvir/Envir.cs
Suprcode: get rid of the if(match) else and and wrap the foreach loops which you only want to be used when if(match) is true

IHaveALongName: Fixed

IHaveALongName:
> Several incomplete features and weird code issues to fix here. No need to rush pull requests through until it 100% works as intended.
> 
> Also please provide screenshots when forms are changed/added so its easier to review

* Can now search items by Name & UniqueID

image

PR #980, opened 2024-07-27
Character List

JevLOMCN:
Added a Character List which will display all Characters.

Before this the process of searching Characters was to either go through them individually using the Account Info Form or the Players Online Tab.

Also has a Filter By Item UID/Name box which allows you to quickly identify the owner of an Item.

image

### DIFF
--- a/Server.MirForms/Account/CharacterInfoForm.Designer.cs
+++ b/Server.MirForms/Account/CharacterInfoForm.Designer.cs
@@ -1,0 +1,165 @@
+﻿namespace Server.Account
+{
+    partial class CharacterInfoForm
+    {
+        /// <summary>
+        /// Required designer variable.
+        /// </summary>
+        private System.ComponentModel.IContainer components = null;
+
+        /// <summary>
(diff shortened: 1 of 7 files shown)


Jinspang:
I don't understand why you are doing this. Are you going to fill all the empty spaces in the Server's form? These functions can be completely added in the Account. I really don't understand why I have to write a form to operate it alone.

JevLOMCN:
> I don't understand why you are doing this. Are you going to fill all the empty spaces in the Server's form? These functions can be completely added in the Account. I really don't understand why I have to write a form to operate it alone.

Explained pretty well in the PR Description

CookieLOM:
All things like this are appreciated @JevLOMCN extra information is never a bad thing and if people aren't grateful that's there issue 😁

JevLOMCN:
> All things like this are appreciated @JevLOMCN extra information is never a bad thing and if people aren't grateful that's there issue 😁

Alot of people get upset when it's not some massive new update.

Just out here trying to make GM lives easier 

#GMLivesMatter

PR #981, opened 2024-07-27
Monsters List

JevLOMCN:
Added Monsters List to SMain (Tab)

image


Suprcode:
This is hugely inefficient. You're looping over every map in the server, then over every cell in that maps, then over every object in that cell. I'd be surprised if it didn't grind the server to halt whilst calculating it all.

I don't really get what benefit this form is giving people? Errors are already output in alternative logs so people can already see and fix any spawn issues.

If you need a current mob count on the map you can just use "MonsterCount" which is already available - but i'd be inclined to just scrap this form due to lack of benefit if provides.

PR #1001, opened 2024-10-29
Trust Merchant Listing

JevLOMCN:
-Added TM List to Player Info Form (Displays only the listing that player has listed)
image

-Added an overall Market Form which displays the entire list of Auctions/Consigns. With filter options.
image
image


Suprcode:
If you can't do anything with this information I'm not really seeing the benefits of it?

The "Market" menu option should also be under Account as its not part of the database.

JevLOMCN:
> If you can't do anything with this information I'm not really seeing the benefits of it?
> 
> The "Market" menu option should also be under Account as its not part of the database.

I plan to add functional buttons in a new PR, these will allow the server owner to “Close the listing and refund the player the item” 

I can move the Market form over to the account tab

Suprcode (changes requested):
Can you also move the market class itself out of database and in to account folder.

JevLOMCN:
image
image
image

Issue #865, opened 2023-10-15
Suggtion: add new button for TM


MousaEqilan:
can we add button to collect all sold at once for main source

MousaEqilan:
anyone was apple to add this ?

damianday:
I can add the server code, if you want to do the client UI  code

PR #1015, opened 2024-11-17
#865 Collect all sold from trust merchant

damianday:


Suprcode:
Does this need new title images adding for the button?

PR #1017, opened 2024-11-17
Compression Fixes #968

### LINKED ISSUE #968: Major - NPC Sale list
Each NPC sale list will update with previous NPC sale list opened. 

image
****

image

damianday:
Move CompressBytes/DecompressBytes to Functions.
Add new packet CompressedPacket.
Make use of compression with NPC goods, reduces packet size much!

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -185,6 +185,7 @@ namespace Client.MirScenes
         public static long NPCTime;
         public static uint NPCID;
         public static float NPCRate;
+        public static PanelType NPCPanelType;
         public static uint DefaultNPCID;
         public static bool HideAddedStoreStats;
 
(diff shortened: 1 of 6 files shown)


Suprcode:
You seem to be compressing a packet inside another packet. I dont understand why you wouldnt just build the compression in to all packets naturally, or have a new class certain packets inherit from to compress.

This method seems a tad messy and complicated to implement should any other packets was compression

damianday:
Yeah it's a bit hacked together, I want the flexibility to be able to send and receive the compressed packet in chunks, leave it with me and I'll improve it tomorrow.

Suprcode:
I would probably add a new class the packet can inherit from which overrides the read and write methods.

Then when someone wants to add a new packet that needs chunks/compression they dont need to do anything but change the packets base class

damianday:
Forgot to mention the other reason it was written this way (sending as multiple packets) was due to the size limitation of 65535, but that could be addressed in a separate commit

damianday:
This is looking a lot simpler now, just set the Compressed to true and it will do the rest.
DecompressPacket / CompressPacket can be overriden to provide custom compressing if required.

PR #1022, opened 2024-11-19
Fix #898

### LINKED ISSUE #898: Expanded storage
You can still use expanded storage after it expires, if you circumvent the client checks, server doesn't check.

damianday:


Suprcode (changes requested):

cantea:
From your last code to the storage right now second storage can`t be used after extended

PR #1024, opened 2024-11-24
Better frame skipping Fix #691

### LINKED ISSUE #691: Latest branch of Crystal has terrible action queue delays
The current client side SkipFrame isn't good enough to keep up with the amount of queued actions from players.

This results in many servers having terrible hit detection for melee players due to the lag build-up.

Double skip frames required based on class to provide a decent player experience.

damianday:
Instead of skipping frames, it will speed up actions.
Frame skipping could only occur when NextAction was valid, which could cause more delays

### DIFF
--- a/Client/MirObjects/Effect.cs
+++ b/Client/MirObjects/Effect.cs
@@ -80,7 +80,7 @@ namespace Client.MirObjects
 
             if (CMain.Time <= NextFrame) return;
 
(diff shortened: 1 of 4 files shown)


damianday:
This still needs some work to do, GameScene CanMove will stop frames from speeding up.

Suprcode:
Went with LCD's approach to improve syncing and performance so closing this ticket.

PR #1031, opened 2024-11-27
Map Search

JevLOMCN:
- Added a Map Search option on the Map Info Form.

image

### DIFF
--- a/Server.MirForms/Database/MapInfoForm.Designer.cs
+++ b/Server.MirForms/Database/MapInfoForm.Designer.cs
@@ -166,7 +166,6 @@ namespace Server
             VisualizerButton = new Button();
             toolTip1 = new ToolTip(components);
             MapSearchButton = new Button();
-            RemoveAllSpawnsButton = new Button();
             MapSearchTextBox = new TextBox();
             MapTabs.SuspendLayout();
             tabPage1.SuspendLayout();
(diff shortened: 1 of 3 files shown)


Suprcode:
the remove all respawns button should be inside the respawns tab (next to the existing remove button)

JevLOMCN:
> the remove all respawns button should be inside the respawns tab (next to the existing remove button)

The Remove all respawns button can work if you select multiple map indexes, I can move if you still want me to

Suprcode:
actually i see your logic just deletes from all selected maps - i don't understand the need to delete respawns in bulk like that though?

Surely if anyone wants to redo spawns, they just delete them from 1 map, and add them back.

JevLOMCN:
> actually i see your logic just deletes from all selected maps - i don't understand the need to delete respawns in bulk like that though?
> 
> Surely if anyone wants to redo spawns, they just delete them from 1 map, and add them back.

I agree that you'd probs not want to delete multiple map's respawn data but the way the code does it is by grabbing the selected index so if multiple are selected then it will do that. Would you like me to move it into the Respawns tab ?

Suprcode:
in that case i don't think the button is needed at all. It seems very niche to want to clear out multiple maps at a time.

JevLOMCN:
Button removed

PR #1035, opened 2024-12-07
Experimental client performance improvement

Lilcooldoode:
Would like a bit of testing on this before been merged. During my testing running 2 clients side by side and running up and down with 1 player both clients stayed quite well in sync

### DIFF
--- a/Client/MirControls/MirControl.cs
+++ b/Client/MirControls/MirControl.cs
@@ -993,7 +993,6 @@ namespace Client.MirControls
         public virtual void Redraw()
         {
             if (Parent != null) Parent.Redraw();
-
         }
 
         #region Font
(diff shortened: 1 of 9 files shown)


damianday:
Is the syncing improvement due to the Skip frame now being > 0 instead of 1 or the Move time change in GameScene? I can't see the dictionary helping much with syncing, but it's probably preferable to use List dictionary http://msdn.microsoft.com/en-us/library/system.collections.specialized.listdictionary%28v=vs.110%29.aspx

Lilcooldoode:
> Is the syncing improvement due to the Skip frame now being > 0 instead of 1 or the Move time change in GameScene?

Combination of both

Suprcode:
Merging as i doubt feedback will be raised otherwise.

PR #1036, opened 2024-12-17
GameShop Form Additions

JevLOMCN:
Added a Button to add an item directly from the Game Shop Form instead of needing to use the ItemInfoFormNew. I've left that one untouched.

Also adds the ability to change an existing GameShop item to any Item from the Item Database.

image


Suprcode:
do you really need to know an items "index" and do all that joining and splitting on the text?

The item name is unique itself so you shouldn't need to show the index with it. I'd just add a list of item names in alphabetical order.

PR #1039, opened 2024-12-30
Change Pet Mode disconnect fix

makivellimir:
Added timer so user only change pet mode once every 0.5 seconds to prevent packet spam disconnect when holding Ctrl+A.


Suprcode:
Closing this PR as another has been opened

PR #1040, opened 2025-01-02
Various

makivellimir:
- Fix to stop Tao pets untaming on logout if Settings.PetSave is set to false.
- Fix for Hero buffs not saving on logout. Note the buffs don't currently save to the DB so are still lost if server is rebooted.
- Fix for change attack mode spam causing disconnect/IP ban due to large number of packets.
- Fix for hero spell usage spam causing disconnect/IP ban due to large number of packets.

### DIFF
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -320,6 +320,12 @@ namespace Server.MirObjects
                                     }
 
                                     break;
+
+                                case (MirClass.Taoist):
+                                    if (pet.Name == Settings.SkeletonName || pet.Name == Settings.AngelName || pet.Name == Settings.ShinsuName)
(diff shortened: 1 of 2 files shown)


Server/MirObjects/PlayerObject.cs
```diff
@@ -320,6 +320,12 @@ public void StopGame(byte reason)
                                     }
 
                                     break;
+
+                                case (MirClass.Taoist):
+                                    if (pet.Name == Settings.AngelName || pet.Name == Settings.ShinsuName)
```
Suprcode, line 325: This is missing the skeleton pet

Suprcode (changes requested):

makivellimir:
Oops - good spot, should be fixed now.

PR #1041, opened 2025-01-03
Client TopMost fix

makivellimir:
Fix for client being set to/remaining 'always on top' when changing from windowed mode -> full-screen -> back to windowed mode.


Suprcode:
You need to merge the last commits before i can merge this.

Recommend in future to create a seperate branch for each of your fixes, as although youre creating multiple pr's it's bringing through all changes

PR #1067, opened 2025-03-03
Update to the Korean version of Legend: Enhanced Position Memory Book…

PMySQLQQ:
… feature

Update to the Korean version of Legend: Enhanced Position Memory Book feature
[Title.zip](https://github.com/user-attachments/files/19054277/Title.zip)

If this get's approved you will need to upload that lib so it can be placed on the patcher
image
image
The above issue has been modified to add the configuration to Server\Debug\Configs.Setup.ini


JevLOMCN:
You’ve removed the Guild Territory Dialog

PR #1071, opened 2025-03-06
GT recall fixes

makivellimir:
- Recall all guild members to GT fixed
- Recall individual guild member to GT fixed


applestroll:
This is an amazing update. I support it!

PR #1074, opened 2025-03-07
Update PlayerObject.cs

ALLFUKING13:
Added CHM effects 1,2,3 lvl's


makivellimir:
The variables you've declared aren't used anywhere so can be removed, also the frames for levels 2 and 3 don't currently exist in the Magic2 lib, so you will need to provide the updated lib file with the new frames.

JevLOMCN:
This isn't from official so it won't be accepted.

PR #1087, opened 2025-05-21
Translate project to Chinese

zoushanji:
This commit introduces Chinese localization for the Legend of Mir 2 project.

Key changes include:

1.  **Shared/Language.cs:** Translated all static string fields to Chinese. This file serves as the primary source for most user-facing text in the client and server. The existing INI loading mechanism will now populate with Chinese strings.

2.  **Client/Forms/AMain.cs & Server.MirForms/SMain.cs:** Refactored hardcoded strings in the client launcher and server control panel UI. These strings were moved to `Shared/Language.cs` and then translated to Chinese.

3.  **PatcherWebSite/mir2-patchsite/index.html:** Translated user-facing text (title, headings, button labels) on the patcher website.

4.  **.resx Files:** A review of a significant portion of the .resx files indicated they are primarily used for resources like icons and images, not for storing user-facing strings. No direct string translations were made in these files based on this pattern.

Further testing by building and running the application is recommended to ensure all translations appear correctly and to catch any remaining untranslated elements.

### DIFF
--- a/Client/Forms/AMain.cs
+++ b/Client/Forms/AMain.cs
@@ -88,13 +88,13 @@ namespace Launcher
             }
             catch (EndOfStreamException ex)
             {
-                MessageBox.Show("End of stream found. Host is likely using a pre version 1.1.0.0 patch system");
+                MessageBox.Show(GameLanguage.Launcher_EndOfStreamError);
                 Completed = true;
                 SaveError(ex.ToString());
             }
             catch (Exception ex)
             {
-                MessageBox.Show(ex.ToString(), "Error");
+                MessageBox.Show(ex.ToString(), GameLanguage.Launcher_ErrorTitle);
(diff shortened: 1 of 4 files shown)


JevLOMCN:
Great addition but now this is forcing Chinese lang by default, should implement a Language option on AMain to allow switching.

afoxsonoco:
This is an AI driven PR. The added variables are welcome, but yes the default language for everything should be kept English.

daneo1989:
Pushing a comit from English to Chinese is pointless when this is an English Master version. 

If you added the ability to have both English and Chinese or multiple languages which can be chosen inside Mir2.ini  🤷‍♂️

Suprcode:
This is an AI driven PR. The added variables are welcome, but yes the default language for everything should be kept English.

PR #1089, opened 2025-05-22
Update GameScene ItemSet_ToolTip

daneo1989:
displaying ItemSet bonuses in the tooltip.

Shows the name of the set

Lists required item types with ✔ or ✘ to indicate which are equipped

Displays bonus stats when the full set is worn

Bonuses appear in white when active, gray when incomplete

Image 0


zhaokai1982:
吉林爱尚智能影音已经查收此邮件，我们保证对所发资料的保密性。有合作意向随时联络。感谢您的配合。

PR #1092, opened 2025-06-01
More HTTP functionality + Allow Observe Config

JevLOMCN:


Suprcode:
Doesn't seem overly secure to allow anyone to request the entire account database with their user names, whether they're an admin, and their gold/credit

PR #1093, opened 2025-06-02
Observe button now displayed

JevLOMCN:
Observe Button is now being displayed on the CharacterInspectWindow. I've moved the bottom buttons left slightly to all the Observe button to fit neatly.

I've set the button to be displayed regardless of if the Character has observe enabled/disabled but this wont bypass the AllowObserve check so is only a visual change.

image


JevLOMCN:
Is [@5b3479f ](https://github.com/Suprcode/Crystal/pull/1093/commits/5b3479f2f2b151b86437c0e392f490a89270cc38) a valid fix?

PR #1106, opened 2025-11-11
Fix Socket Exception Error

phillipgibbings:
When an IP address is configured but is invalid or unavailable, the server will throw a SocketException error. To resolve this issue, the system will display a prompt asking whether you would like to bind the server to 0.0.0.0 (also known as IPAddress.Any). Selecting this option allows the server to start and operate normally by accepting connections on all available network interfaces.

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -1,4 +1,4 @@
-﻿using Client.MirControls;
+using Client.MirControls;
 using Client.MirGraphics;
 using Client.MirNetwork;
 using Client.MirObjects;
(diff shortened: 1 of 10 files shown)


JevLOMCN:
Will this save the IP to the setup.ini when utilised?

phillipgibbings:
Add admin-only Benediction Oil control with step-based luck/curse and global shout

Implements an admin-only prompt when using Benediction Oil for controlled luck or curse adjustments.
Adds confirmation flow for adding Luck (+1, capped at +7) or Curse (+1 per step, up to C+10)
Broadcasts admin actions via global shout (ChatType.Shout2)
Keeps standard RNG behavior unchanged for normal players
Each admin action consumes exactly one oil

PR #1110, opened 2025-11-23
Bug Fixes, Localization improvements

TiomTang:
Bug Fixes:
1. Added Level property for Monster and passed it to the client.
2. Added DisplayMonsterLevel configuration option in Settings.
3. Fixed an issue where quest items could not be correctly obtained.
4. Fixed a regex error in ItemData.
5. Removed the redundant NameLabel.Text = Name.
6. Fixed the misaligned display of AMain in the client on Windows 11.

Localization Feature Updates:
1. Improved and refined the original localization feature, adding a large amount of text. Using JSON format for storage eliminates the need to define numerous string constants, which was both prone to omission and cumbersome when saving and retrieving file content.
2. Introduced a new external database localization feature to facilitate bulk localization using tools, improving efficiency and avoiding data inconsistencies caused by database modifications. Particularly when localizing items, extensive edits to drop table files were previously required, which were error-prone and involved significant workload. Usage: After the server starts, a DbLanguage.json file will be generated in the root directory. Localizing this file and placing it in the client’s root directory will complete the localization of a large volume of database text without any impact on the database.

The motivation behind improving the localization feature was my experience during database localization: after modifying item names, the server would encounter drop loading errors. Facing the massive number of drop table files was overwhelming, so I took the initiative to enhance and refine this localization feature.

### DIFF
--- a/Client/MirScenes/Dialogs/HeroDialogs.cs
+++ b/Client/MirScenes/Dialogs/HeroDialogs.cs
@@ -771,7 +771,7 @@ namespace Client.MirScenes.Dialogs
                     Library = Libraries.Prguse,
                     Parent = this,
                     Sound = SoundList.ButtonA,
-                    Hint = string.Format(GameLanguage.ClientTextMap[nameof(ClientTextKeys.HeroBehaviourFormat)], Enum.GetName(typeof(HeroBehaviour), i)),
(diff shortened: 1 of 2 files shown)


JevLOMCN:
You're still doing it, I can do the or for you if needed

TiomTang:
@JevLOMCN Sorry, I'm not very familiar with this. I would be very grateful for your help if you could.

Suprcode:
Pull request is way too big. Happy to accept fixes, but they must be 1 feature at a time in managable chunks we can review.

PR #1111, opened 2025-11-23
Improve drawing performance

TiomTang:
1. DrawFloor optimization, compared to the pre-optimization performance, the time consumption has been reduced by approximately 56.2% after statistics.
2. Texture cleaning algorithm optimization
3. Lighting processing

Before：
image

After：
image


Suprcode:
Pull request is way too big. Happy to accept fixes, but they must be 1 feature at a time in managable chunks we can review.

PR #1114, opened 2025-11-25
Comprehensive Map-Based Restrictions & System Behaviour

daneo1989:
Group Restrictions (NoGroup)
• Map logic now enforces group rules immediately across movement, login, teleport, and group changes. • If a party has two or fewer members, the group auto-disbands on entry. • If a party has three or more members, only the entering player is removed. • RequiredGroup enforcement now applies instantly with no grace period. • All affected players receive contextual chat messages.

Pet Restrictions (NoPets)
• Pets no longer die on entering restricted maps.
• They are instead cleanly frozen: targets cleared and PMode = None. • Visual feedback is sent via ObjectTurn broadcasts. • Pets automatically unfreeze when returning to allowed maps. • Removed outdated/inapplicable references (CanMove, CanAttack, Enqueue, Visible).

Hero Restrictions (NoHero)
• Heroes are safely and automatically despawned when entering restricted maps. • Players receive clear chat notifications.

Experience Restrictions (NoExp)
• EXP gain is fully blocked on maps flagged with NoExp. • EXP validation updated to properly respect map.Info.NoExp.

FireWall Count & Limit System
• Introduced caster-bound FireWall caps to control concurrent FireWalls. • FireWall counters cleanly reset on expiration or logout. • Ensures predictable and safe spell behaviour across all map types.

Map-Entry Handling & Safety Improvements
• Unified NoGroup / NoHero / NoPets processing on map entry. • Map-change cleanup reorganized to avoid safe-zone bounce behaviour and null-reference exceptions. • Added atomic DisbandGroup() to prevent list modification during iteration. • Centralized pet freeze/unfreeze logic using existing Frozen + PMode systems. • All changes tested and validated with no regressions.

### DIFF
--- a/Client/MirGraphics/DXManager.cs
+++ b/Client/MirGraphics/DXManager.cs
@@ -519,7 +519,7 @@ namespace Client.MirGraphics
                 }
 
                 DXManager.FloorTexture = null;
-                    GameScene.Scene.MapControl.FloorValid = false;
+                GameScene.Scene.MapControl.FloorValid = false;
 
                 if (DXManager.FloorSurface != null && !DXManager.FloorSurface.Disposed)
                 {


daneo1989:
Map Restrictions

JevLOMCN:
Does the No Pet option include/exclude pick up pets? I'd personally have this as a separate option considering they are different types of pets.

Server/MirObjects/HumanObject.cs
```diff
@@ -2462,11 +2465,18 @@ public bool Walk(MirDirection dir)
             Direction = dir;
             if (CheckMovement(location)) return false;
 
-            CurrentMap.GetCell(CurrentLocation).Remove(this);
+            var srcCell = CurrentMap.GetCell(CurrentLocation);
+            if (srcCell?.Objects != null && srcCell.Objects.Contains(this))
```
Suprcode: Unnessessary checks here

Server/MirEnvir/Map.cs
```diff
@@ -1103,6 +1103,9 @@ private void CompleteMagic(IList<object> data)
                     value = (int)data[2];
                     location = (Point)data[3];
 
+                    int castId = 0;
+                    if (data.Count >= 5 && data[data.Count - 1] is int ci) castId = ci;
```
Suprcode: You don't want to be selecting the last data item, if someone passes through more data items this would break. Check the count but then use the correct index that the castId will always be passed in.

Server/MirObjects/PlayerObject.cs
```diff
@@ -382,7 +398,23 @@ public void StopGame(byte reason)
             }
 
             Envir.Players.Remove(this);
-            CurrentMap.RemoveObject(this);
+
+            try
+            {
```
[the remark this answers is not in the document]
Suprcode: The same issue still stands. Not sure why CurrentMap / validpoint/cell/object/contains are all being checked here.

None of this should be possible so these checks are redundant. What was wrong with the existing code as it was?

Server/MirObjects/PlayerObject.cs
```diff
@@ -380,7 +396,16 @@ public void StopGame(byte reason)
             }
 
             Envir.Players.Remove(this);
-            CurrentMap.RemoveObject(this);
+
+            if (CurrentMap != null && CurrentMap.ValidPoint(CurrentLocation))
+            {
+                var cell = CurrentMap.GetCell(CurrentLocation);
+                if (cell?.Objects != null && cell.Objects.Contains(this))
```
Suprcode: Recommend to revert this change back to how it was. None of these things should be possible at this stage, so its just more error catching.

Suprcode (changes requested):
Aplogies, i thought i sent this comment for review a few days ago but didn't click submit.

daneo1989:
resolved conflict as per request.

Server/MirObjects/NPC/NPCSegment.cs
```diff
@@ -236,11 +236,6 @@ public void ParseCheck(string line)
                     CheckList.Add(new NPCChecks(CheckType.CheckMap, parts[1]));
                     break;
 
-                case "MAPLIGHT":
```
Suprcode: Looks like this PR hasn't brought through the previous map light additions. This needs pulling in otherwise it'll merge it out.

PR #1116, opened 2025-11-26
Implement configurable Monster Rarity & Recall systems

daneo1989:
Rarity:
- Added shared MonsterType enum and MonsterRarityData helper with per-tier profiles (HP/DEF/DMG/EXP/GOLD multipliers, drop bonuses, display colour).
- Monsters roll a rarity type at spawn (unless disabled) and adjust stats, EXP, gold drops, and displayed name/colour accordingly.
- Rarity settings now live in Configs/MonsterRarity.ini (Enabled flag, tier spawn odds, multipliers/bonuses). Settings.Load reads the file and injects the values into the helper; Setup.ini no longer recreates those keys.

Recall:
- Introduced a CanRecall flag on MonsterInfo (editable in MonsterInfoFormNew), allowing specific monsters to teleport back to their target when kited.
- Added global recall controls in Setup.ini (MonsterRecallEnabled, MonsterRecallRange, MonsterRecallCooldown) so behaviour can be tuned or disabled server-wide.
- MonsterObject honours both the per-monster flag and the global settings; recalls now play the teleport effect for better visual feedback.

This uses Version = 115 to suit to follow after Map Restrictions update.


Server/Settings.cs
```diff
@@ -1582,5 +1618,98 @@ public static void SaveGoods()
...
+                    ItemDropBonusPercent = MonsterRarityEliteItemDropBonusPercent,
+                    GoldDropBonusPercent = MonsterRarityEliteGoldDropBonusPercent,
+                    NameColour = MonsterRarityData.GetProfile(MonsterType.Elite).NameColour
+                }
+            };
+
+            MonsterRarityData.ConfigureProfiles(overrides);
+        }
+
+        private static double ReadRarityChancePercent(InIReader reader, string percentKey, string legacyKey, double defaultPercent)
```
Suprcode: Don't understand legacy key?

PR #1117, opened 2025-11-26
Enhanced Localization Support

TiomTang:
1. Expanded the original localization system. All client‑side and server‑side text entries have now been added to the local Language.json file.

2. Added an external (overlay) localization mechanism for database text. When the server starts, it generates a DbLanguage.json file. You can localize all database‑related entries by editing this file with any suitable tool, without making any changes to the actual database.

   The motivation for implementing this feature was that while localizing item data, I found that I also needed to modify the files storing monster drop information, which would have been a large amount of work. This external database localization feature solves that problem.

### DIFF
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -2054,7 +2054,7 @@ namespace Server.MirObjects
 
                             if (player == null)
                             {
-                                ReceiveChat(string.Format(GameLanguage.ServerTextMap.GetLocalization(ServerTextKeys.CouldNotFind), parts[0]), ChatType.System);
+                                ReceiveChat(string.Format(GameLanguage.ServerTextMap.GetLocalization(ServerTextKeys.CouldNotFind), parts[1]), ChatType.System);
                                 return;
                             }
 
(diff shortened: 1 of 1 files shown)


JevLOMCN:
This would make having multiple client language options easier, maybe with a language option in Mir2Config.ini which loads on client start and then picks the correct language entries inside Language.json

JevLOMCN:
I'm not talking about changing the server ui language but the client language as you've moved most hard coded text from the code into its own language file

TiomTang:
I share your opinion; the UI I mentioned refers to the client. I have tried Zircon Mir3, and in the game you can switch to the appropriate localization files by setting the language.

Suprcode:
Accessing the array directly and having to use "nameof" for every lookup is quite messy (and error prone as you have no guarantee the item is in the dictionary). It would be better to just call a method and then you can easily add further support and checks without having to change the code again.

Instead of 

`GameLanguage.ClientTextMap[nameof(ClientTextKeys.PriceGold)]`

You would do

`GameLanguage.ClientTextMap.GetText(ClientTextKeys.PriceGold)`

Using a simple method like below

```
public static string GetText(this IDictionary<string,string> map, ClientTextKeys key)
{
    return map[key.ToString()];
}
```

Suprcode:
The huge static dictionarys also don't seem great and well managed. If theres any change in the localisation system it would be better to use Microsofts officially supported resx system rather than implementing another custom system.

https://learn.microsoft.com/en-us/dotnet/core/extensions/localization

PR #1118, opened 2025-11-26
Bug fix

daneo1989:
GM CMD @Revive message fix, Before and After output message.
Image 4
Image 5

### DIFF
--- a/Server/MirObjects/PlayerObject.cs
+++ b/Server/MirObjects/PlayerObject.cs
@@ -2082,6 +2082,7 @@ namespace Server.MirObjects
                                 switch (ob.Race)
                                 {
                                     case ObjectType.Player:
+                                    case ObjectType.Monster:
                                         if (ob.Dead) continue;
                                         ob.EXPOwner = this;
                                         ob.ExpireTime = Envir.Time + MonsterObject.EXPOwnerDelay;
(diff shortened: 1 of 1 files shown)


Server/MirObjects/PlayerObject.cs
```diff
@@ -3971,7 +3984,7 @@ public void Chat(string message, List<ChatItem> linkedItems = null)
                             ReceiveChat(string.Format("You have deleted skill {0} from player {1}", skill1.ToString(), player.Name), ChatType.Hint);
                             player.ReceiveChat(string.Format("{0} has been removed from you.", skill1), ChatType.Hint);
 
-                            Helpers.ChatSystem.SystemMessage(chatMessage: $"{player} Skill {skill1.ToString()} was removed by GM: {Name}");
+                            Helpers.ChatSystem.SystemMessage(chatMessage: $"{player.Name} Skill {skill1.ToString()} was removed by GM: {Name}");
```
Suprcode: does "chatMessage:" need to be removed here like the above one?

daneo1989: Once you @DeleteSkill it would display Server.MirObjects.PlayerObjects Skill FireWall was removed by GM: X

@KILL didn't display anything. It has the kill message but never work, So whether we want this? or remove.

Server/MirObjects/PlayerObject.cs
```diff
@@ -2082,12 +2082,25 @@ public void Chat(string message, List<ChatItem> linkedItems = null)
                                 switch (ob.Race)
                                 {
                                     case ObjectType.Player:
+                                        if (ob.Dead) continue;
+                                        ob.EXPOwner = this;
+                                        ob.ExpireTime = Envir.Time + MonsterObject.EXPOwnerDelay;
+                                        ob.Die();
+
+                                        if (ob is PlayerObject killedPlayer)
```
Suprcode: Check is redundant here. ob.Race already confirmed the ob is a player

Suprcode (changes requested):

Suprcode (approved):

PR #1119, opened 2025-11-26
Hide zero values in tooltips for duration/weight

daneo1989:
Item tooltips now suppress redundant “0” entries. Duration-specific text (Usage, Loyalty, Durability, etc.) is generated only when the current—and where relevant, maximum—durability values are greater than zero, so you no longer see strings like “Usage 0/0” or “Loyalty 0/0”. Likewise, the weight snippet is appended only when `HoverItem.Weight` is positive. If both fields would have been zero, that entire tail section stays hidden, keeping tooltips concise and focused on meaningful data.
Image 6

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -6731,19 +6731,24 @@ namespace Client.MirScenes
                 switch (HoverItem.Info.Type)
                 {
                     case ItemType.Amulet:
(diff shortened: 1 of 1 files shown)


Suprcode:
Not accepting this for now as it'll conflict with the language PR. Once language PR has been accepted this will need updating to use new language strings.

PR #1123, opened 2025-11-28
Fix : correct NameLabel color update behavior

TiomTang:
When checking `LabelList[i].ForeColour != NameColour`, the code incorrectly created a new `MirLabel`, causing Taoist/Wizard pets to fail to update their name color after leveling up.

When an existing NameLabel with the same name is found, its properties should be updated instead of creating a new instance.

### DIFF
--- a/Client/MirObjects/HeroObject.cs
+++ b/Client/MirObjects/HeroObject.cs
@@ -48,13 +48,16 @@ namespace Client.MirObjects
 
             for (int i = 0; i < LabelList.Count; i++)
             {
-                if (LabelList[i].Text != ownerText || LabelList[i].ForeColour != NameColour) continue;
+                if (LabelList[i].Text != ownerText) continue;
                 OwnerLabel = LabelList[i];
                 break;
             }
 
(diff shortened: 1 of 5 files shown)


Suprcode:
Yeah I get that 1 name label could be used for 2 monsters, and draw the first monster in white, then get the same label and update its colour to red and draw that.

However every time you change the color you're sending a change event that any number of systems can subscribe to - so if you're constantly doing draw updates of one label and switching back and forth between colours you're constantly calling the "OnColorChanged" event.

Not only will that be slower, but could cause side affects if something needs to know when a labels color has changed.

I would add a test message to your OnForeColourChanged event - i suspect when you have 2 monsters sharing the same labels with different colours that you're going to be spammed with the event calls on every redraw.

TiomTang:
If that’s the case, I don’t think using the name property as the ID for a MirLabel object is a good idea. A MirLabel with the same name might represent different states of the same monster (different colors, shown at different times), or it might be used by two completely different monsters. This can easily lead to confusion. Introducing an objectID as the identifying key would probably be a more appropriate solution.

PR #1124, opened 2025-11-28
Npc map light

daneo1989:
[@MAIN]
#IF
MAPLIGHT DAWN
#ACT
GOTO @GREET_DAWN
#ELSEACT
GOTO @CHECK_DAY

[@CHECK_DAY]
#IF
MAPLIGHT DAY
#ACT
GOTO @GREET_DAY
#ELSEACT
GOTO @CHECK_EVENING

[@CHECK_EVENING]
#IF
MAPLIGHT EVENING
#ACT
GOTO @GREET_EVENING
#ELSEACT
GOTO @CHECK_NIGHT

[@CHECK_NIGHT]
#IF
MAPLIGHT NIGHT
#ACT
GOTO @GREET_NIGHT
#ELSEACT
GOTO @GREET_FALLBACK

<$MAPLIGHT> will display current map light cycle.

### DIFF
--- a/Server/MirObjects/NPC/NPCSegment.cs
+++ b/Server/MirObjects/NPC/NPCSegment.cs
@@ -1736,7 +1736,9 @@ namespace Server.MirObjects
                         break;
 
                     case CheckType.CheckMapLight:
-                        failed = !IsMapLightConditionMet(param);
+                        var globalLight = Envir.Lights.ToString().ToUpperInvariant();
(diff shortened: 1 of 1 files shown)


Suprcode:
Yes the way its been done is still hugely over complicated (and it still has the unnecessary ANY logic). The entire check can be done on a single line like most other check commands.

Instead of being comma separated it would probably be better to just accept a single time of day, and use the npc's OR system if you want multiples. This keeps it in line with every existing check command.

Use "CheckDay" for your logic

PR #1131, opened 2025-12-04
Safe zone walkthrough

daneo1989:
- Added `AllowSafeZonePassThrough` setting (default true) to server config, persisted in INI.
- Exposed the toggle in `ConfigForm` under Game World so admins can enable/disable it.
- Server walk/run logic now checks the toggle before letting players/pets pass through each other in safe zones.


Suprcode:
Strange logic throughout here. We already have an "InSafeZone" check server side, and the check should probably be implemented on the "Blocking" variable. I'm surprised this even works

Issue #1078, opened 2025-03-11
Why is the textbox masking incorrectly like this?


jamir2:
![Image](https://github.com/user-attachments/assets/3c1609b0-e0a6-474c-9e37-3e63cdfa3dd2)
Almost every place where textbox is used does this

Suprcode:
Yes this is a known issue with textboxes

daneo1989:
Not sure if its something todo with Window mode, but when full screen it doesnt happen.

<img width="1024" height="768" alt="Image" src="https://github.com/user-attachments/assets/f06908b7-f77d-4140-9442-b24d6e83d4e2" />

phillipgibbings:
i have submitted a fix for this, @Suprcode, can you test and verify and if your happy accept the pull request

<img width="1031" height="806" alt="Image" src="https://github.com/user-attachments/assets/34b9607a-e6fa-4cc3-97b4-9c55425386ac" />

https://github.com/Suprcode/Crystal/pull/1182

PR #1142, opened 2025-12-06
#1078 TextBox masking in window mode.

daneo1989:
- `UpdateTextBoxHostLocation` now centralizes where the WinForms input control lives. In fullscreen it sits where it always was; in windowed mode it’s moved to `HiddenTextBoxLocation` so it never paints over Mir UIs.
- Mouse input paths (`OnMouseDown/Move/Up/DoubleClick/Wheel`) detect when the control is hidden and replay the corresponding Windows messages via `SendMessage`, using `CMain.MPoint` to compute the caret position. Keyboard input still flows through the real `TextBox`, so all selection/editing behavior is preserved without rewriting it manually.
- Texture creation is still triggered everywhere (windowed + fullscreen), so the Mir-themed mask/caret drawn from `TextBox.DrawToBitmap` now matches in both modes.
2image
1image

@Suprcode not sure whether this is the correct direction which AI provided.

### DIFF
--- a/Client/MirControls/MirTextBox.cs
+++ b/Client/MirControls/MirTextBox.cs
@@ -78,6 +78,31 @@ namespace Client.MirControls
                 TextBox.Location = DisplayLocation;
                 _textBoxOffscreen = false;
             }
+
+            UpdateMouseMoveHook(!_textBoxOffscreen);
+        }
+
+        private void UpdateMouseMoveHook(bool enabled)
+        {
+            if (TextBox == null || TextBox.IsDisposed)
+                return;
+
+            if (enabled)
+            {
+                if (_mouseMoveHooked)
+                    return;
+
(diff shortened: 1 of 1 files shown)


Suprcode:
Your screenshot doesn't show any text in the text boxes, or active carats.

Show screenshots of the textbox in edit mode with a full page of text, and another of the chat bar with actife carat

shellohunter:
https://github.com/Suprcode/Crystal/blob/627e11f4f083e27a206b2953a1c9e103d823cb1d/Client/MirNetwork/Network.cs#L41

I noticed another issue with textbox: in some scenarios client tries to update gui from network thread, which is not a good idea, and runtime will raise an exception by default.
this happens for all mir controls with an embedded winform control.

ref: https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/how-to-make-thread-safe-calls

Issue #966, opened 2024-06-06
Fishing Rod / Mount Socket information.


daneo1989:
Fishing Rod showing, Socket information even though sockets 0 on Database, 

Fishing equipment can only be right-clicked into slot and not click and dropped. 

Possibly remove "Socket" and change to something suiting and remove Fishing Menu (N)? 

Same issue with Mounts
![Untitled](https://github.com/Suprcode/Crystal/assets/6216097/37fd1dee-5764-4399-b05e-5d44b4937aed)

Suprcode:
Fishing and mounts have fixed socket counts as they have a UI dependant on them.

PR #1145, opened 2025-12-06
Fishing Rod / Mount Socket information. #966

daneo1989:
Hide socket tooltip rows for fishing rods and mounts so their hover info no longer mentions sockets or the socket dialog tip
Image 7
Image 8


daneo1989:
While the socket system is complete custom. I plan on submitting some big changes expanding on the system.
 
While you say it’s good information to have. You can clearly open the socket window and equip items without having to open the Fishing Dialog and Mount Dialog. Sockets also have a problem is that Socketed items don’t add towards the current items stats it’s hidden. So currently linking items only displays the Socketed item name not the stat. While this commit doesn’t stop opening that socket dialog it basically hides the information. 

My plan with the socket system.

Issue #803, opened 2023-04-24
HP + MP %


daneo1989:
Does not allow for HP + MP % Recovery.

PR #1157, opened 2025-12-18
HP + MP % #803 HP/MP potion ranges + % recovery (editor + tooltip + i…

daneo1989:
…ndicators)

Adds proper potion recovery support for both fixed values and ranges.

- **ItemInfo / DB**
  - Added `HPRollRaw` / `MPRollRaw` to store HP/MP ranges like `10~100`.
  - Implemented `RollHP()` / `RollMP()` helpers.
  - Saved ranges using an appended tail marker and load-time “peek + rewind” so older DBs still read cleanly.

- **Item editor (ItemInfoFormNew)**
  - `StatHP` / `StatMP` now accept either an integer or `min~max`.
  - Ranges are persisted to `HPRollRaw/MPRollRaw`, fixed values stay in `Stats[HP/MP]`.
  - Validation added with a clear “enter a number or min~max” message.

- **Server potion logic**
  - NormalPotion (shape 0) now supports range/% by rolling into the potion pools.
  - SunPotion (shape 1) supports range/% and shows indicators for both HP and MP when applied.
  - Refactored the repeated math into a single helper on `HumanObject` to keep player/hero behavior identical.
  - MP regen ticks now emit indicators the same way HP regen already did.

- **Client tooltip + indicators**
  - Potion tooltips show “Health/Mana Recovery + …” with ranges, and show percent recovery as “Health/Mana Recovery + X%”.
  - Added `HealthRecoveryPlusPercent` / `ManaRecoveryPlusPercent` localization keys (English + Chinese).
  - DamageIndicator display now formats positive values as `+X` (keeps the old color scheme) and skips `0` spam.


Suprcode:
I can't even review this PR for fixes. Its so all over the place that it sounds like you're just asking AI to do something and then instantly accepting what it gives you.

Why would you have a stat column allow a string of min and max, when we have specific stats of min and max it could use?

Suprcode:
What i mean is we have plenty of stats available to use. If normal recovery only uses the HP/MP stat currently you could just reuse an existing stat which isn't used. Or add a new "MaxHP" stat that it can use if this is going to be a common thing. The orbs for example just use other stats for their own use.

Hacking it in and using custom "HPRollRaw" strings outside of the stat system is just a crazy way to do it.

You really need to check what the AI is providing you

Issue #755, opened 2023-04-05
Hero Buffs


JevLOMCN:
Buffs like HealthWine can be used on heroes however expire upon removing and summoning the hero. (Needs to save it's time left)

PR #1162, opened 2025-12-20
#755 Hero Buffs

daneo1989:
Tried to replicate proposed issue, Could not replicate.

If you shut down the server it would remove the Hero Buffs. Sorry @Suprcode included some more changes to this commit.

1. Hero buffs acts like Players buff on server restart/shutdown. (Retains the buffs)
2. Hero buff border to follow Official (Red border)
3. Hero Name to follow Official (Purple)
4. If you seal the Hero it will now remove the active buffs.

image


Suprcode (changes requested):
Strange image logic needs correcting

PR #1163, opened 2025-12-23
Monster Rarity enhancement.

daneo1989:
Rarity Colour's can now be changed inside the MonsterRarity.ini

 Default is now set False.
 
MonsterInfoFormNew IsBoss bool now restricts from spawning enhanced rarity.


Suprcode (changes requested):
Wrong color translator used.

This whole system would be cleaner if it didn't have hard coded variables for each rarity and instead just had a multiplier for the default values. 

It already does this mostly for the profiles, it just needs the hard coded variables for the defaults removing, and instead for each rarity just increase by a certain amount. Once the defaults are set it then just reads from the config anyway.

PR #1165, opened 2026-01-22
Add XML documentation for NPCAction enum

mikepattyn:
Added XML documentation for all NPCActions
Generated with Cursor. 

Gives us a nice tooltip with params and example's on hover
image
image

### DIFF
--- a/Server/MirObjects/NPC/NPCActions.cs
+++ b/Server/MirObjects/NPC/NPCActions.cs
@@ -1,4 +1,4 @@
-﻿namespace Server.MirObjects
+namespace Server.MirObjects
 {
     public class NPCActions
     {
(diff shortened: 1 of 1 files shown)


JevLOMCN:
The example you provided is incorrect, mapfilename not mapname

So it would be move 0 100 100

Full list of NPC commands exist here with a description https://www.lomcn.net/wiki/index.php/NPC_Commands

Suprcode:
Personally don't see the benefit here. Usually you add summary info to help the developer on how to use the variables/methods. But in this case the functionality of all these commands is already done, so a developer has no need to use them for anything else?

This is wiki info that doesn't belong in a code file imo

PR #1170, opened 2026-02-02
Quest Icons as Per KR.

daneo1989:
Version updated to 118

Door icons are drawn on the world map, mini‑map, and big map. Multi‑tile doors are clustered and the icon is centered. Quest icon priority is based on quest type (Story > General > Daily > Repeatable), then completion state, then id. Per‑quest flags in `QuestInfoForm` control whether a quest’s icon shows on door world/mini/big map (and NPC mini/big map).

Image 88
Image 93
Image 87


Suprcode:
looks weird having npc quest icons on doors - is that what they do on official now?

daneo1989:
Images699
Images700

PR #1171, opened 2026-02-02
Item tooltip KR Style update.

daneo1989:
Item tool tip KR style update,

Wasn't able to keep original Durability location without conflicting with Item Type & Weight. color would change for Type,weight. If i split these up placement just makes it look weird. So now moved onto its own line.

Examples

Durability: 24/25(25 -1)

when Min is added: DC + 3 ~ 17(2 ~ 15 + 1 ~ 2)
when only Max is added: DC + 2 ~ 17(15 + 2)

Majority of Item enhancements are done with Max DC,SC,MC,AC,MAC but if you take Sockets into count you could potentially upgrade Min.

Color:
- Cyan when any added value is positive
- Red when any added value is negative
- White when no added
Images711

Image 100
Image 98

### DIFF
--- a/Client/MirScenes/GameScene.cs
+++ b/Client/MirScenes/GameScene.cs
@@ -1,4 +1,4 @@
-﻿using Client.MirControls;
+using Client.MirControls;
 using Client.MirGraphics;
 using Client.MirNetwork;
 using Client.MirObjects;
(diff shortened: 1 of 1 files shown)


xD-E-A-T-Hx:
Yuck. Adds too much noise imo. 
Could it not be done if you press example ctrl and hover over to show what's added or removed.

daneo1989:
> Yuck. Adds too much noise imo. Could it not be done if you press example ctrl and hover over to show what's added or removed.

This is how official has it displayed, But it does offer Ctrl + Left Click to show/hide Set Info

PR #1177, opened 2026-05-21
ADD AOI System in Server

liaojiaao:
16X16 grid AOI  to replace orgin 17X17 grids search .
17X17 times search get objects. now only need 2X2 times get objects.

Maybe it can improve performance.
Need test and compare old version. (maybe i missed something)
I just one player in my local notebook

### DIFF
--- a/Server.MirForms/Server.csproj
+++ b/Server.MirForms/Server.csproj
@@ -29,7 +29,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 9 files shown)


Suprcode:
Original system uses a Globals.DataRange, yours is hard coded to 15.

I don't think efficiency will be improved, you're still doing the same loop over all objects + you're now also doing a linq query so performance might even go down.

